PR #26, opened 2016-06-28
Initial refToHTML implementation

jure:
First go (and no styling, I leave this to you, styling experts) :)

What we perhaps could do is to assemble an in-between CSL JSON, and use something like Citeproc.js to convert to HTML. That might be a bit more robust still.

image


michael:
Thanks for that! Mergin... 

Will try to also use this in RefTarget component (see #29 )

michael:
Good job. Activated also for ref target editing dialog.

image

PR #29, opened 2016-06-28
Xref editing

michael:


packages/jats/xref/XRefTargets.js
```diff
@@ -1,31 +1,71 @@
...
+
+    // Update model
+    var newTargets = node.targets;
+    if (newTargets.indexOf(targetNode.id) > 0) {
+      newTargets = without(newTargets, targetNode.id);
+    } else {
+      newTargets.push(targetNode.id);
+    }
+
+    // TODO: this is causing the modal to close. Can we prevent that?
```
michael: @oliver---- you have an idea why the modal closes here? Can't figure out what's going on..

michael:
@oliver---- Pls review and merge. There's one bigger glitch related to the modal closing as you toggle reference targets. I wanted to discuss a some other UX options but we can do that in a separate iteration.

The basic UX principle so far is that if you open the ref target editor the selected targets are shown on top. As you select other targets they will be where they are until you close the dialog and reopen (then they'll appear on top again).

oliver----:
Need to take a close look.

PR #45, opened 2016-07-03
Move front/body/back editors into components.

michael:
WIP - don't merge!


michael:
@oliver---- I created components for Article, Front, Body, Back and instantiate them using the registry. I would approach the author by overriding certain components. E.g. Front would do some cherry picking instead of rendering the whole front container as done in publisher mode. ok with that approach?

We enter some competing scenarios with our configurator api's. I'd like to start addressing this as well, so we can have author+reader+publisher packaged in one app. See https://github.com/substance/substance/issues/734

oliver----:
Fine with me. I can't say yet which is the ideal way. So let's go ahead and see where we get.

PR #54, opened 2016-07-07
Use two configuration levels.

michael:
On top level we configure the Scientist. The ScientistConfigurator then
contains separate configurators for each mode (PublisherConfigurator,
AuthorConfigurator).


michael:
First iteration on having two-level configuration. Only jats-editor example ported yet. But I have a good feeling with that approach so far. It's very modular, and I didn't have to change any of the core API's.

Any objections using the top level configurator to manage sub configurators (one for each available mode: publisher, author, reader,...).

```js
'use strict';

var AuthorPackage = require('../author/package');
var AuthorConfigurator = require('../author/AuthorConfigurator');
var PublisherPackage = require('../publisher/package');
var PublisherConfigurator = require('../publisher/PublisherConfigurator');

module.exports = {
  name: 'scientist',
  configure: function(config) {
    // Default configuration for available scientist modes
    config.addConfigurator('author', new AuthorConfigurator(AuthorPackage));
    config.addConfigurator('publisher', new PublisherConfigurator(PublisherPackage));
  }
};
```

michael:
Please review. Ready to merge if you are ok with the changes.

oliver----:
I don't see why the subclasses of ScientistConfigurator are necessary. The only difference is the package.

packages/scientist/package.js
```diff
@@ -0,0 +1,15 @@
...
+var AuthorPackage = require('../author/package');
+var AuthorConfigurator = require('../author/AuthorConfigurator');
+var PublisherPackage = require('../publisher/package');
+var PublisherConfigurator = require('../publisher/PublisherConfigurator');
+
+module.exports = {
+  name: 'scientist',
+  configure: function(config) {
+    // Default configuration for available scientist modes
+    config.addConfigurator('author', new AuthorConfigurator(AuthorPackage));
```
oliver----: You could use `new ScientistConfigurator(AuthorPackage)`.

oliver----:
Let's merge this.

PR #62, opened 2016-07-12
Add support for article-meta element.

michael:


michael:
@oliver---- pls review and merge. I think it's ok for a start but we should discuss some possible strategies for editing of contrib elements, which will vary for Author and Publisher.

PR #103, opened 2016-10-07
Add strict mode

naveenarun:
I was having trouble with strict mode (node make -s -w) on Node version 4.4.7 on OSX. I added "use strict" in make.js to explicitly let the "let" statements run.


michael:
I've removed the ES6 syntax so you should now able to run in Node 4. Thanks for reporting.

PR #125, opened 2017-02-24
WIP: XSD driven editing capabilities

oliver----:
We want
1. to be really JATS compliant when importing XML
2. to be able to only create JATS compliant content upfront (as opposed to a-posteriori validation)
3. to be able to validate when we need
4. an XSD driven implementation to reduce amount of implementation


oliver----:
The following image reveals the mindful reader that we can already create a Finite State Machine (DFA) for each element that will help us with validation plus prediction.
 
image

What you see is a graph representation of a DFA storing element tagNames on outgoing edges between nodes that represent states in a sequence of valid elements.
eg: `{ 1: { 2: ['sec-meta', 'Symbol(EPSILON)'] } }` means that to get from state `1` to state `2`, either 'sec-meta' must be presented or the empty word (EPSILON) -- in other word: at this very position in the schema 'sec-meta' is an optional element.

oliver----:
@michael subs-dancing ;)

michael:
Schweet! :)

PR #126, opened 2017-03-05
add index + download files

PookMook:


PookMook:
woops, my bad, wrong repo for my test (sorry!)

PR #144, opened 2017-06-17
Multiselect

michael:


michael:
Hey looking good. Few things:

- Please don't change package.json (substance#multiselect), because then I can't merge PR (just use npm link for local setup)
- It would be better to emit a `change` event for the multiselect rather than listening to click

Safe to merge after those changes

Integral:
No problems, but still we should use refs to get values. Pls merge.

Issue #152, opened 2017-06-19
Translations


michael:
Here's the UI for title and abstract translations.

![image](https://user-images.githubusercontent.com/2931/27288350-c7a684d2-5506-11e7-807e-fd6bed4fd606.png)

Integral:
How it will look like from schema point of view?

michael:
I created a setup for you (translations branch)

Note, that I changed the JATS model a bit and make it more explicit.
https://github.com/substance/texture/commit/d7e1711560b3532885b526b0e1ddb54fbd08d9df


Title translations go into `trans-title-group`: 

https://github.com/substance/texture/blob/translations/data/kitchen-sink.xml#L21

Abstract translations go into `trans-abstract-group`:

https://github.com/substance/texture/blob/translations/data/kitchen-sink.xml#L86

PR #161, opened 2017-06-21
Translations

Integral:
As described in #152.
There is still a question what we should we orphaned nodes, will it be deleted automatically?
If I removing abstract-content or group, should I also remove each paragraph?
P.S.: Languages are defined here: https://github.com/substance/texture/blob/translations/src/editor/components/TranslationsComponent.js#L3


michael:
I will manually merge this, as it conflicts with my panel navigation work.

michael:
has been merged manually.

Issue #164, opened 2017-06-22
Improvements list


Integral:
- [x] style menu according mockup
- [x] fix contributors meta section bugs and missing contributor name input borders
- [x] if there is no affiliations you should see warning message "Please add affiliations..." in contributors panel
- [x] add placeholders for translation and other inputs 
- [x] whole meta panel must be grey
- [x] implement references removing
- [x] use trash icon anywhere for removing instead of x
- [x] append footnotes list after references list (with remove and add functionality)

PR #165, opened 2017-06-22
Hot improvements

michael:


Integral:
@michael what we discussed in #164 is done I would say.
However there is still a few problems:
1) #166 
2) When you did delete a reference or footnote it is somehow got cached in edit xref tool, so you can still choose there what you already delete. Of course it would lead to errors.

You will also need to recompile schema as I marked fn as a container in TextureJATS.classification.json

michael:
👍 well done!

PR #231, opened 2017-08-03
WIP: Ability to insert a <disp-quote> element.

michael:


michael:
@NateWr won't get to this today. Hopefully tomorrow.

michael:
Insertion works fine now! However, performing an undo right after insertion causes an error. Looks like the XRefComponent gets rerendered while the node has actually been deleted already. I figure the XRefComponent has not been disposed properly (unregister from labels:generated event)?

michael:
Stacktrace: 
image

michael:
@oliver---- I implemented InsertDispQuoteCommand. However, the validator claims that `<disp-quote-content>` is not allowed inside `<disp-quote>` and rejects the operation, while this should be perfectly valid. 

To reproduce the error: Click Insert -> Blockquote

The problem is that `schema.findLastValidPos(this, child.type)` returns -1 instead of 0. Might be an edge case that the DFA misses.

image

oliver----:
There is no problem with dispose. But what you do there is to emit labels:generated during the 'update' stage, where XRefComponent is calling rerender right away.
Thus, rerender is called too earlier, particularly before the TextProperty has been rerendered.

oliver----:
The problem with the insert is, that it chokes with the partially invalid element (`<attrib> is not there yet, but mandatory)
I need to rethink this, as we only need to make sure things are valid at the end of a transaction.

src/article/TextureEditingInterface.js
```diff
@@ -17,21 +17,4 @@ export default class TextureEditingInterface extends EditingInterface {
     return this.getDocument().createElement(...args)
   }
 
-  /*
-    2.0 API suggestion (pass only id, not data)
-  */
-  insertBlockNode(nodeId) {
```
oliver----: @michael I think we want to go for a purely Node/Element oriented interface. Using the `id` I consider more low-level and thus more 1.0 style.

michael: You are right. I like that API!

oliver----:
We are now validating changed elements after a transaction.

oliver----:
@michael u can continue

michael:
That looks great. Thanks! Think we are done with this PR. Are we good to publish new Substance preview and merge?

oliver----:
Probably I need to improve the XML API further when working on Stencila, but I understand that we want to get this here out for further testing.

michael:
Should we continue with beta7-preview.x or start with 1.0.0-preview1? :)

oliver----:
let's do it ;)

michael:
@oliver---- i will do it if ok?

PR #261, opened 2017-09-08
Use Button component everywhere.

Integral:
@michael pls review


michael:
Looking good. Will run it and merge a bit later.

PR #263, opened 2017-09-10
Fix empty ref creation.

Integral:
@michael this is fix for adding and editing new reference


michael:
Oh that was a bad mistake. Thanks!

PR #264, opened 2017-09-13
Allow empty fn-group in TextureJATS.

Integral:


src/article/TextureJATS.rng
```diff
@@ -756,7 +756,16 @@
...
   </define>
 
-
+  <!--
+    Allow empty <fn-group>
+  -->
+  <define name="fn-group-model">
+    <zeroOrMore>
+      <choice>
+        <ref name="fn-link.class"/>
```
michael: There is no point in using a a class `fn-link.class `here, as there is only one element inside it. Better to do <ref name="fn">, that's more readable.

Issue #256, opened 2017-09-05
whitescreening on disp-quote elements when contained within footnotes


axfelix:
Hi guys,

At first I thought we might be producing some invalid JATS, but this should actually be OK (which explains why this is getting past your validator but still crashing Texture).

Basically, it looks like Word for Mac (and only Word for Mac) occasionally formats footnotes such that when they're run through our parsing engine, the resulting JATS comes out like:

`<fn><p><disp-quote><p>Content</p></disp-quote></p></fn>`

This is pretty ugly but actually *should* be OK as far as I can tell. Let me know if you need a doc to reproduce with (not sure if I can share @jalperin's manuscript).

michael:
Hi Alex! Can you provide a minimal test file, containing nothing but the content that causes the error? Just so we can reproduce, then we can address it.

axfelix:
[28-99Z_Article Text-93-1-2-20170906.zip](https://github.com/substance/texture/files/1281764/28-99Z_Article.Text-93-1-2-20170906.zip)

Sure, this should do it

michael:
This comes from a normalization step we have taken. We do a transformation to pull out block elements when wrapped inside a <p> element. We do this to pull out fig elements from paragraphs and move them to the top level for instance.

In your footnote case it turn this:

```xml
<fn id="bibd2e184">
  <p>
    <disp-quote>
      <p>foo</p>
    </disp-quote>
  </p>
</fn>
```

into this:

```xml
<fn id="bibd2e184">
  <disp-quote>
    <p>foo</p>
  </disp-quote>
</fn>
```

And this then fails the JATS4R validation, as disp-quote elements are not allowed as childs of fn. It would also not be valid in JATS. Generally I think this paragraph trick opens a backdoor for a lot of things which then are valid in JATS but not intended.

Do we really want to allow disp-quotes in footnotes.. or would you rather tweak your parsing engine to output just a paragraph instead which would be more footnote-ish?

axfelix:
Hey Michael,

We can look at gradually moving away from this behaviour, but for now there are lots of moving parts involved in doing so unless we just write a sanity check at the end of our process, and while I can do that, I'd rather Texture not be whitescreening on technically valid JATS :)

oliver----:
I agree that it would be great if Texture could deal with all the technically possible variants of content. However, this is technically not feasible. We find restrictions which still give us enough expressivity.
[...]

PR #269, opened 2017-09-16
Footnote normalize transformer.

Integral:
As discussed in #256.


src/article/JATS4R.rng
```diff
@@ -315,6 +315,14 @@
     </element>
   </define>
 
+  <define name="fn">
+    <element name="fn">
+      <oneOrMore>
+        <ref name="just-para.class"/>
```
michael: Also here please use the explicit tag instead of just-para.class, for better readability.

```
<ref name="p"/>
```

test/fixture/fn-group.xml
```diff
@@ -9,6 +9,14 @@
       <fn id="fn2">
         <p>Phasellus ex nulla, interdum condimentum pretium ac, auctor et tellus. Maecenas eu est a neque vestibulum laoreet. Mauris libero nulla, pretium eu iaculis id, mollis ac erat. Suspendisse rutrum gravida malesuada</p>
       </fn>
+      <fn id="fn3">
```
michael: I dont't see a test case checking the fn3 example.

michael:
Typo found in filename: Please rename NormilizeFn.test.js to NormalizeFn.test.js.

Integral:
Pardon me.

michael:
Tests are correct so far, but could be split into separate tests instead of checking everything in one. Better do "test normalize disp-quote wrapped in p" and "test normalize element-citation > disp-quote".

Example with element-citation is actually unlikely to happen (and maybe not a good test) as that would be invalid JATS. If you have time, please see if there's a better example that would pass JATS but should be turned into a p element too. or stripped completely.

For example someone could do this silly thing:

```xml
<fn>
  <p><fig><label>Fig 1</label><caption><title>bla</title><p>the caption text</p></caption><graphic></fig></p>
</fn>
```

The only valid thing you could do here is dropping everything except for the p elements in the structure. Like `fn.find('p p')` - find all nested p's and pull them out to the top level:

```xml
<fn>
  <p>the caption text</p>
</fn>
```

If no nested p's are found in the weird structure then and p must be created e.g.

```xml
<fn>
  <p><graphic/></p>
</fn>
```
to

```xml
<fn>
  <p></p>
</fn>
```

Integral:
That's what transformer doing already, but in other way.
My idea was to test that there that footnote didn't contain any other things except paragraph, e.g. there is no `disp-quote`, `<element-citation>` there. However you are definitely right that it should be valid JATS, I will do this with your example, thx.

PR #290, opened 2017-11-15
Import entitites (References) - Round 1

oliver----:


test/converter/ImportEntities.test.js
```diff
@@ -41,6 +39,58 @@ test("Import journal citation", function(t) {
   t.end()
 })
 
+test("Import book citation", function(t) {
+  let dom = DefaultDOMElement.parseXML(fixture)
+  let entityDb = _emptyEntityDb()
+  let api = _createAPI(entityDb)
+  let converter = new ImportEntities()
```
oliver----: You could put these four lines into a helper, to avoid redundancy

test/converter/ImportEntities.test.js
```diff
@@ -13,12 +13,8 @@ const fixture = readFixture('element-citation.xml')
 
 
 test("Import journal citation", function(t) {
-  let dom = DefaultDOMElement.parseXML(fixture)
   let entityDb = _emptyEntityDb()
-  let api = _createAPI(entityDb)
-  let converter = new ImportEntities()
-
-  converter.import(dom, api)
+  _importEntities(entityDb)
```
oliver----: You can improve that, by returning an object and using spread expression to assign multiple args
```
let { entityDB, converter } = _setup(fixture) 
```

oliver----:
We should enable the `make test` by adding a dependency to `test:node`

PR #293, opened 2017-11-16
Export entities.

Integral:


oliver----:
Looking good. Could you rebase onto master.
Then we will have CI :)

PR #304, opened 2017-11-23
Connect editor with new entity-based reference editing

michael:


src/converter/r2t/ConvertReproFig.js
```diff
@@ -31,23 +31,23 @@ export default class ConvertReproFig {
...
-    //   reproFig.attr('fig-type', 'repro-fig')
-    //   reproFig.append(
-    //     $$('alternatives').append(
-    //       exportSourceCode(source),
-    //       exportOutput(output)
-    //     )
-    //   )
-    // })
+    reproFigs.forEach((/*reproFig*/) => {
+      throw new Error('Revise implementation')
```
oliver----: 👍 Thanks

PR #329, opened 2017-12-10
Drag and drop relationships.

Integral:


Integral:
@michael pls review

src/util/updateEntityChildArray.js
```diff
@@ -1,22 +1,21 @@
-import { without } from 'substance'
-
 export default function updateEntityChildArray(editorSession, nodeId, tagName, attribute, oldEntityIds, newEntityIds) {
 
   editorSession.transaction(tx => {
     let node = tx.get(nodeId)
 
-    let addedEntityIds = without(newEntityIds, ...oldEntityIds)
-    let removedEntityIds = without(oldEntityIds, ...newEntityIds)
+    // TODO: do we still want to calculate the difference here?
```
michael: If you put such todo's, please leave the original code, otherwise we won't understand what calculate the difference means.

PR #332, opened 2017-12-13
Entity Importers

michael:


michael:
@oliver---- I worked on entity importer/exporters, as we discussed. Maybe you could have a look to check if on the right track. I'll be in Amsterdam for next few days and will return on Tuesday.

Code is not functional yet, just a sketch.

michael:
Code is now functional, and supports journal and book citations. Remaining pub types must be added and tested.

michael:
@Integral can you have a look at the code and give me feedback? You could take over the task of implementing the remaining pub types if you want to.

Integral:
@michael, are you sure that PatentNumber and PatentCountry are not an array [here](https://github.com/substance/texture/blob/master/src/entities/EntityDatabase.js#L78)? E.g. are you sure that i can't use few `patent` elements inside patent publication type reference?

michael:
We need to check with the publishers. But I think it would be good to have them as single properties, otherwise the UI gets complicated (create, delete buttons etc.). I will review your PR today and get this stuff merged.

PR #339, opened 2018-01-16
Insert figures.

Integral:
Insert image blobs via tools and drag and drop.


Integral:
There is file resolve issue in travis ci, however nothing like this exists locally and all tests are green.

michael:
That was breaking because you used `InsertFigure.js` (uppercase) in the fileName and `insertFigure` (lowercase) in the import. So it works only on case-insensitive file systems. Added a fix.

Integral:
Thanks.

PR #363, opened 2018-02-02
Refactor.

Integral:


src/util/removeElementAndXrefs.js
```diff
@@ -0,0 +1,23 @@
+import { without } from 'substance'
+
+export default function removeElementAndXrefs(editorSession, elementId, parentElName) {
+  let doc = editorSession.getDocument()
+  let xrefIndex = doc.getIndex('xrefs')
+  let xrefs = xrefIndex.get(elementId)
+
+  if (xrefs.length === 0 || window.confirm(`If you delete this reference, it will also be removed from ${xrefs.length} citations. Are you sure?`)) { // eslint-disable-line
+    editorSession.transaction(tx => {
+      let parentEl = doc.find(parentElName)
```
michael: This could be dangerous if there's more than one of these tagnames in the document. Maybe better pass in the parentEl node directly, so the caller can make sure the right element is taken. Otherwise looking good! 👍

michael:
👍 Merging.

PR #366, opened 2018-02-05
Setup image upload based on archive abstraction

michael:


src/VfsClient.js
```diff
@@ -33,15 +33,17 @@ function _readRawArchive(fs, darUrl) {
...
       data: content
     }
   })
   assets.forEach(asset => {
     let path = asset.attr('path')
     rawArchive[path] = {
+      encoding: 'url',
       type: 'image/jpg',
-      url: path
+      data: './data/kitchen-sink/'+path
```
Integral: are you sure?

michael: Well, this is just for the dev setup, where we only have the kitchen sink example. ;) But yeah, if we wanted more, we need to change this.

PR #390, opened 2018-02-17
Added email attribute to contrib.

Integral:
Email attribute is currently missing inside contrib element.
This PR adds email to JATS4M schema, person entity schema and entity converters.


michael:
👍

Issue #374, opened 2018-02-08
Improve UI flow for adding cross references


michael:
Feedback by eLife: Overall, there seems to be a lot of double actions required - e.g. adding a citation of a reference does not open a list of the available references, but inserts a placeholder citation that then needs to be clicked to select the right reference. This could get quite frustrating if writing/editing a long article.

michael:
@Integral I think we'd need to change `InsertXrefCommand` or `InsertInlineNodeCommand` to select the newly inserted inline node. then the dialog should pop up immediately and we don't need an extra click.

PR #396, opened 2018-02-22
Improve UI flow for adding cross references.

Integral:
Why: double actions required when inserting new cite
What: auto open reference selector when inserting new cite as proposed in #374


michael:
Merci.

PR #397, opened 2018-02-23
Multiple publisher name and location.

Integral:
There are books published in two different locations, or by two Publishers.
This PR changed the current schema to allow using publisher-name and publisher-loc more than one time per ref.

### DIFF
--- a/src/converter/r2t/EntityConverters.js
+++ b/src/converter/r2t/EntityConverters.js
@@ -834,16 +834,12 @@ function _createTextElement($$, text, tagName, attrs) {
 function _createMultipleTextElements($$, text, tagName, attrs) {
   if (text) {
     const textItems = text.split(';')
-    let elements = []
-    textItems.forEach(textItem => {
-      // Removing a white space at the start of text
-      // in case of separation with semicolon and space
-      const text = textItem.replace(/^\s+/g, "")
-      const el = $$(tagName).append(text)
+    const elements = textItems.map(ti => {
+      const el = $$(tagName).append(ti.trim())
       forEach(attrs, (value, key) => {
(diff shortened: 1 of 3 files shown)


src/converter/r2t/EntityConverters.js
```diff
@@ -794,6 +794,15 @@ function _getText(rootEl, selector) {
   }
 }
 
+function _getSeparatedText(rootEl, selector) {
+  let match = rootEl.findAll(selector)
+  if (match) {
+    return match.map(m => { return m.textContent }).join('; ')
```
michael: you can write even shorter `.map(m => m.textContent)`.

src/converter/r2t/EntityConverters.js
```diff
@@ -822,6 +831,21 @@ function _createTextElement($$, text, tagName, attrs) {
   }
 }
 
+function _createMultipleTextElements($$, text, tagName, attrs) {
+  if (text) {
+    const textItems = text.split('; ')
```
michael: If someone writes "foo;bar" without the space, it would not split it. So I guess here it would be better to split by ';' and then trim the whitespace for each item.

Integral:
Agree

michael:
@Integral I will now merge some stuff in Substance/DarServer/Texture. Will ping you when done, please rebase before you add your fixes here.

michael:
@Integral woops forgot to commit something. Will move back into office to get to the code (give me 1 hour)

michael:
Good to merge @Integral. Just one more thing before you do: 

Please update the dar spec (publisher-loc?, publisher-name? entries should turn into publisher-loc*, publisher-name*)

Issue #427, opened 2018-03-12
Drag and Drop: Handle situations when a file is dropped outside the dropzone


michael:
Currently in Texture Desktop the image replaces the web-view contents.

michael:
Fixed via #429

PR #429, opened 2018-03-12
Supress global dnd events on app level

Integral:
As described in #427 we have a problems when a file is dropped outside the dropzone. In Desktop app it replaces a web-view, in browser it will open a local file. 
To fix that we should supress dnd events on app level. This PR do exactly that for a browser and desktop versions.


michael:
Merci

Issue #433, opened 2018-03-13
Can't install npm dependency from github


de-code:
Hi,

It would be nice if it was possible to install the dependency from github since it deviated quite a bit from the last release. e.g.:

```bash
npm install github:substance/texture --save
```

That 'installs' fine but the dependency can't be imported. Which I think is because the bundle is missing.

michael:
I've added a postinstall hook, to build the bundle. Can you try again?

michael:
Well that does not seem to work. I think we need the npm releases because of that said build step.

de-code:
Okay, I think some other projects either provide a bundle built by CI or the source is standard enough to be understood. I noticed that your latest preview is actually in npm. So it is probably not an issue if those are available frequently.

michael:
Yes, we'll count them up frequently (at least the preview channel), as we have softwares depend on Texture. E.g. https://github.com/stencila/stencila

oliver----:
@de-code I think this should be possible now, after I have re-factored the build configuration via #454.

PR #454, opened 2018-03-23
Fix package configuration

oliver----:
## Why?

Currently Texture can not be used as a node module anymore.
Reason is, that the main entry is pointing to the electron main entry.

## What?

Fix respective entries in package.json


oliver----:
@michael plz check, merge and publish

oliver----:
the cjs build is missing. Hmmm, why was this removed?

oliver----:
Changed the npm scripts and make targets to what IMO makes most sense.
@michael plz check

michael:
The problem is we have kind of a hybrid situation now, with Texture being an Electron application and a library. Maybe we should aim for two different package.json files, the one that goes into dist could be the one for electron, the other for the classic library.

I also had to move "dar-server" from dev dependencies into dependencies, because the electron app needs it, but the library does not.

Maybe we can run a build step that generates a patched package.json and adapting it for electron?

Oh and where is the cjs build currently used?

oliver----:
Pls take a look at my latest changes - if this is serving what you need.

You already have an electron package.json in the dist folder, which I think is fine.
The root package.json should be the 'library' one.
Texture is used as a library e.g. by nodejs/coverage tests in Stencila.

michael:
Thanks, I'll check in an hour if that's ok?

oliver----:
Yep. I am working with a local install currently.

michael:
App packaging (npm run release) seems to rely on the top level package.json, investigating if it can be changed to dist/package.json.

oliver----:
I think this is relevant: https://www.electron.build/tutorials/two-package-structure

michael:
I can't figure it out. :(

It's too intransparent too me what happens behind the scences in the two package json structure mode. The build step fails with this error:

Error: Application entry file "main.js" does not exist. Seems like a wrong configuration.
    at /Users/michael/projects/substance/texture/node_modules/electron-builder-lib/src/platformPackager.ts:349:15

However main.js is present in app/main.js as well as in dist/main.js. I assume for some reason the electron builder is looking for it in "."   - like it reads the correct file (from app/package.json) which has "main": "main.js" but doesn't look for it in 'dist'.

Want to take a look? Can commit what i have here.

oliver----:
I take a look.

I think the most straight-forward is to keep the regular dist bundle simple, i.e. without electron.
And instead replicate 'dist' into 'app/dist'. Let electron be run there, and 'electron-builder' too.

michael:
Wait just investigating one more thing. The config build/files is relative to the app directory, so maybe need to fiddle there.

You know how to run builder manually without the npm release command?

oliver----:
`node_modules/.bin/builder`

michael:
Ok, seems you can't break out of the app folder in the config. Tried:

```js
    "files": [
      "*", // copy stuff from app folder - works
      "../dist/**/*", // this does not work
      "node_modules/**/*" // works
    ],
```

Arr, these hybrid setups are really tricky.

oliver----:
Let us do what I proposed. The current way is only working, because we do not use native node modules. I suppose it would fail e.g. with something like gitlib or sqlite.
For that we must run `install-app-deps`. And then I think it is better to create a `app-dist` where everything is just plain and simple.

michael:
Pushed what I have, please take over! :)

oliver----:
Are u ok with my proposal?

michael:
Yes, we don't need native modules, but I have nothing against the two package structure so we can be more explicit with electron app related config and library config. Still the top level package.json contains stuff of both worlds...

Regarding replicating into app/dist? would this be a copy or would use generally change dist to app/dist ?

oliver----:
> Yes, we don't need native modules

Not yet. But maybe when we start doing more heavy stuff like real-time collab and the git-based versioning.

> Regarding replicating into app/dist?

Copying, of course. Otherwise we have the same problem just changed the folder name ;)

michael:
But wait, if we can replicate / copy the app folder into dist (which we already do) then we can make this the "app" folder that builder needs. 

Let me try..

oliver----:
Then you gonna publish everthing to npm.. Let's not.

oliver----:
I am pretty convinced that what I propose is the cleanest.
Because it separates the 'library' version from the 'app' version.
No tricks. No confusion.

michael:
Hmm.. when publishing to npm, we wouldn't run electron builder right? But yeah, it would spoil the dist folder with node_modules etc.

michael:
Fine with your suggestion. It's just we'll have duplicate generated files (texture.js et. al), that fine?

Want to take over?

oliver----:
What is the problem with these being duplicate?

michael:
No it's fine, as you say these are just two separate bundles. Just slightly more 'work' for "node make".

oliver----:
It is fine if 'node make' is doing the work, instead of us.

michael:
With a build step for app we could also change main.js and app.js to ES6.

oliver----:
... and generate `package.json`

michael:
Regarding splitting things, we may want to have two folders now:

`web` -  with index.html + editor.js   compiles to dist/
`app` - with app.html + app.js     compiles to app/dist

oliver----:
Is `editor.js` not relying on electron?
Then it could go into the regular dist.

oliver----:
Ah, you want to have a pure web version, that we can copy on a server.

michael:
No editor.js is only relevant for the web version. There's a lot of duplicate code between app.js and editor.js - I'm planning to come up with an AppChrome baseclass to do this better. Also to share code between Stencila.app and Texture.app.

oliver----:
Ok. I agree it should be part of the regular dist.
`editor.js` is not used for the electron build at all?

michael:
Correct.

oliver----:
Could you go through the npm scripts and think about which make target should be used. I.e. thinking about which scripts are actually necessary, and used by whom for what.
Then we can call the most specific make targets and keep build times low.

michael:
- `start` (should only build the web distro) and serve it
- `app` (should only build the app/electron distro) - ideally we could let this watch for changes too, so we can just reload in electron after making changes; wasn't able to make this work as initially node make app should be run and then electron ./app/dist ?
- `release` does a clean app build and runs electron builder
- the build and watch scripts be removed i think

oliver----:
Thx.

oliver----:
@michael should the 'data' folder be in the app dist?

michael:
Yeah we need it to read the data/blank document.

oliver----:
@michael plz try out

michael:
Looking good. Liking how it's separate and more transparent!

npm run release works, except the app icon is not considered anymore in the new build (the files are in the build folder on top level)

michael:
Pushed a fix for the build resources. Please review.

oliver----:
BTW, #433 should be possible now.

oliver----:
Merging.

PR #465, opened 2018-04-04
Import bibliography.

Integral:
Why: We want to give an ability to import ref by providing DOI list and fetching data.
What: This PR adds an ability to enter list of DOIs and fetch data via DataCite.

### DIFF
--- a/src/converter/bib/BibConversion.js
+++ b/src/converter/bib/BibConversion.js
@@ -1,0 +1,60 @@
+/*
+  Converts a CSLJSON record to our internal format.
+  See EntityDatabase for schemas.
+*/
+
+export function convertCSLJSON(source) {
+  let bibType = source.type
+  let result
+
+  if (bibType === 'article-journal') {
+    result = _convertJournalArticleFromCSLJSON(source)
+  } else {
+    throw new Error(`Bib type ${bibType} not yet supported`)
+  }
+  return result
+}
+
+function _convertJournalArticleFromCSLJSON(source) {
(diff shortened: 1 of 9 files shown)


Integral:
@michael couple of thoughts: we should check no connection case and fetch data from doi.org (for that we need special util, so we can do it with custom headers).

michael:
@Integral took a quick look. 

Few bugs:

- authors and editors are not extracted currently
- split DOIs by space (`" "`) not by `","`

Missing:

- error handling and reporting
  - convertCSLJSON should throw an error for unsupported types (and try catch outside)
  - errors (like loading errors or conversion errors) should be indicated and displayed
- loading indicator (and blocking ui while loading)

UI with loading and error states could look like this:

image

michael:
Will do a quick round of code cleanup, expect some commits in next 30minutes.

Integral:
Let's decide what exactly do we need until merge and improve other things
afterwards

michael:
Yep. When there are some invalid DOI's, we should not update the search string. Just leave everything as is. The user then can fix it manually (e.g. when they made just a typo).

michael:
Need a bit more time. Will write to you in Slack soon.

Integral:
@michael I think we should merge now. Then Philipp could work on other converters and I will work more on:
- no connection errors
- doi.org parsing
- import CSL JSON via DND

michael:
Thank you @Integral! I will check and merge asap.

PR #468, opened 2018-04-07
Import bib improvements.

Integral:
Why: To improve bibliography import workflow we decided:
1) Fetch CSL JSON from doi.org
2) Upload CSL JSON via drag and drop or upload dialog
3) Check internet connection status
What: This PR contains 1-2 items.

### DIFF
--- a/src/converter/bib/BibConversion.js
+++ b/src/converter/bib/BibConversion.js
@@ -39,6 +39,10 @@ function _convertJournalArticleFromCSLJSON(source) {
     }
   })
 
+  if(!data.doi) {
+    throw new Error(`Citation must have DOI.`)
+  }
+
   return data
 }
 
--- a/src/editor/components/AddReferenceComponent.js
+++ b/src/editor/components/AddReferenceComponent.js
@@ -1,6 +1,6 @@
 import { Component } from 'substance'
 import DOIInputComponent from './DOIInputComponent'
-//import ReferenceUploadComponent from './ReferenceUploadComponent'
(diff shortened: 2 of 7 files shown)


Integral:
@michael here we are... I think we need to polish error reporting for json upload.
Any thoughts regarding internet connection status? Maybe we should override substance sendRequest util to track http error status?

michael:
Nice work @Integral!

I think we don't need to do some special checking of the internet connection. Just somehow transport the error after you click on the "Add" button.

I did some testing and these things should be fixed:

- currently you can insert the same DOI twice (maybe we can use the DOI as an id, to make it unique)
- Some items are skipped when importing from an uploaded json file  (see
[Exported Items.json.zip](https://github.com/substance/texture/files/1887233/Exported.Items.json.zip) which contains a report type)
   - should produce an error instead and import nothing
   - means that all items need to be converted first and only if there is no error send the action up so all the items get added to the bibliography
- I think errors should be displayed inline as you had it before (display below the dropzone or below the search field)

Can add those fixes to the current PR? Merci :)

michael:
Hey thanks. Let me know when done and i will check and mrge.. :)

Integral:
@michael it's done now. Couple of things to notice:
1) Last commit fixes citation label recalculation which I found again
2) If you will try to import CSL with already imported DOI it will not stop the import, e.g. it is not the error, but this record will be skiped
3) If you will try to import CSL with citation without DOI it will show you the error (i have an example of journal article without DOI from Zotero).

michael:
Great!

Issue #431, opened 2018-03-12
Import of bibliography in CSL JSON


michael:
E.g. we should be able to read:

```js
[
	{
		"id": "http://zotero.org/users/local/V2QRG5IH/items/JNF5KMDK",
		"type": "article-journal",
		"title": "Object vision to hand action in macaque parietal, premotor, and motor cortices",
		"container-title": "eLife",
		"volume": "5",
		"source": "CrossRef",
		"URL": "https://elifesciences.org/articles/15278",
		"DOI": "10.7554/eLife.15278",
		"ISSN": "2050-084X",
		"language": "en",
		"author": [
			{
				"family": "Schaffelhofer",
				"given": "Stefan"
			},
			{
				"family": "Scherberger",
				"given": "Hansjörg"
			}
		],
		"issued": {
			"date-parts": [
				[
					"2016",
					7,
					26
				]
			]
		},
		"accessed": {
			"date-parts": [
				[
					"2018",
					3,
					12
				]
			]
		}
	}
]
```

and map accordingly to our internal schema:

```js
JournalArticle.schema = {
  type: 'journal-article',
  authors: { type: ['person'], default: [] },
  editors: { type: ['person'], default: [] },
  articleTitle: { type: 'text', optional: true },
  source: { type: 'string', optional: true },
  volume: { type: 'string', optional: true },
  issue: { type: 'string', optional: true },
  year: { type: 'string', optional: true },
  month: { type: 'string', optional: true },
  day: { type: 'string', optional: true },
  fpage: { type: 'string', optional: true },
  lpage: { type: 'string', optional: true },
  pageRange: { type: 'string', optional: true },
  elocationId: { type: 'string', optional: true },
  doi: { type: 'string', optional: true }
}
```

More publication types here:

https://github.com/substance/texture/blob/master/src/entities/EntityDatabase.js 

I will do a first shot and then get help from @zuphilip.

michael:
We'll start working on support to import CSL/JSON records. 

@Integral in order to fetch recoreds from DataCite we can use the following url:


https://data.datacite.org/application/vnd.citationstyles.csl+json/DOI

E.g. 
https://data.datacite.org/application/vnd.citationstyles.csl+json/10.5063/f1m61h5x

michael:
New Modal shown after clicking "Add Reference(s)".

![image](https://user-images.githubusercontent.com/2931/38162532-f00fe320-34e2-11e8-8fc5-edf44365a160.png)

[...]

PR #480, opened 2018-04-15
Extend mapping from CSL JSON

zuphilip:
See #431

### DIFF
--- a/src/converter/bib/BibConversion.js
+++ b/src/converter/bib/BibConversion.js
@@ -7,30 +7,121 @@ export function convertCSLJSON(source) {
   let bibType = source.type
   let result
 
-  if (bibType === 'article-journal') {
-    result = _convertJournalArticleFromCSLJSON(source)
+  // CSL types: http://docs.citationstyles.org/en/stable/specification.html#appendix-iii-types
+  let typeMapping = {
+    "article": "journal-article",
+    "article-magazine": "journal-article",
+    "article-newspaper": "journal-article",
+    "article-journal": "journal-article",
+    //"bill"
+    "book": "book",
+    //"broadcast"
+    "chapter": "book",
+    "dataset": "data-publication",
+    //"entry"
+    "entry-dictionary": "book",
+    "entry-encyclopedia": "book",
+    //"figure"
+    //"graphic"
+    //"interview"
+    //"legislation"
+    //"legal_case"
+    //"manuscript"
+    //"map"
+    //"motion_picture"
+    //"musical_score"
+    //"pamphlet"
+    "paper-conference": "conference-proceeding",
+    "patent": "patent",
+    //"post"
+    //"post-weblog"
+    //"personal_communication"
+    "report": "report",
+    //"review"
+    //"review-book"
+    //"song"
+    //"speech"
+    "thesis": "thesis",
+    //"treaty"
+    "webpage": "webpage"
+    //NA : "periodical"
+    //NA : "clinical-trial"
+    //NA : "preprint"
+    //NA : "software"
+  }
+
+  if (typeMapping[bibType]) {
+    result = _convertFromCSLJSON(source, typeMapping[bibType])
   } else {
     throw new Error(`Bib type ${bibType} not yet supported`)
   }
   return result
 }
 
-function _convertJournalArticleFromCSLJSON(source) {
+function _convertFromCSLJSON(source, type) {
   const date = _extractDateFromCSLJSON(source)
 
   let data = {
-    type: 'journal-article',
-    authors: source.author.map(a => {return {surname: a.family, givenNames: a.given}}),
-    articleTitle: source.title,
-    source: source['container-title'],
+    type: type,
+
     volume: source.volume,
     issue: source.issue,
     pageRange: source.page,
     doi: source.DOI,
+    pmid: source.PMID,
+
+    edition: source.edition,
+    publisherLoc: source['publisher-place'],
+    publisherName: source.publisher,
+    pageCount: source['number-of-pages'],
+    isbn: source.ISBN,
+
     year: date.year,
     month: date.month,
-    day: date.day
+    day: date.day,
+
+    uri: source.URL,
+    version: source.version
+
+    /* Examples with no corresponding field:
+        - abstract
+        - accessed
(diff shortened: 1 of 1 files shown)


src/converter/bib/BibConversion.js
```diff
@@ -40,6 +131,8 @@ function _convertJournalArticleFromCSLJSON(source) {
   })
 
   if(!data.doi) {
+    // TODO: We should not rely that the imported item has a DOI, because it can also be imported from a generic CSL JSON file.
+    //  However, there are some problems in the further processing withouth a DOI at the moment...
     throw new Error(`Citation must have DOI.`)
```
zuphilip: @michael Can you help for entries from CSL JSON without a DOI?

src/converter/bib/BibConversion.js
```diff
@@ -7,31 +7,122 @@ export function convertCSLJSON(source) {
...
+    data.title = source.title
+  } else {
+    data.source = source['container-title']
+    if (source.type==='chapter') {
+      data.chapterTitle = source.title
+    } else if (source.type==='dataset') {
+      data.dataTitle = source.title
+    } else {
+      data.articleTitle = source.title
+    }
```
zuphilip: Is there any reason for all these different labels for the title? I.e. `title`, `chapterTitle`, `dataTitle`, `articleTitle` for different item types. It would be much easier to have only one title and a possible container title...

src/converter/bib/BibConversion.js
```diff
@@ -7,31 +7,122 @@ export function convertCSLJSON(source) {
...
+    //"personal_communication"
+    "report": "report",
+    //"review"
+    //"review-book"
+    //"song"
+    //"speech"
+    "thesis": "thesis",
+    //"treaty"
+    "webpage": "webpage"
+    //NA : "periodical"
```
zuphilip: What is the difference between `periodical` and `journal-article`? What would be an example?

zuphilip:
The CSL import works now for all resources with a mapping to the internal item type but only when they have a DOI. Can we change the behaviour, such that also for resources without a DOI this works? Most books don't have a DOI, but there are also journal articles without a DOI.

Most of the mapping is now independent of the item type in order to reduce code duplication. However, it is possible that some fields are now saved for an item type which actually is not possible there. Is this a problem? I guess we could also add another cleaning step which somehow go through the model in `Entity Database.js`, but for that I would need some help how to import this file...

Some more comments also about the reference model is given above.

src/converter/bib/BibConversion.js
michael: Agree we could normalise that in our internal model (EntityDatabse.js) and just call it `title` (needs adjustment of the importer/exporter code).

What do you mean by container title?

src/converter/bib/BibConversion.js
michael: Like a magazine that occurs regularly. But also not sure how it is actually used in practice. I will start an issue to trigger our partners, which should provide real world examples for each pub type.

src/converter/bib/BibConversion.js
michael: This is good for now, but yes we need to think of a solution for this soon. The problem is, that we don't have an easy way to detect duplicates if there's no doi. We'd need to compare if all properties are the same to skip an entry.

src/converter/bib/BibConversion.js
```diff
@@ -7,31 +7,122 @@ export function convertCSLJSON(source) {
...
+  // CSL types: http://docs.citationstyles.org/en/stable/specification.html#appendix-iii-types
+  let typeMapping = {
+    "article": "journal-article",
+    "article-magazine": "journal-article",
+    "article-newspaper": "journal-article",
+    "article-journal": "journal-article",
+    //"bill"
+    "book": "book",
+    //"broadcast"
+    "chapter": "book",
```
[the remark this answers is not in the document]
michael: We could call the label `Book/Chapter` in the edit dialog?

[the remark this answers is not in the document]
michael: Will do that, we can adjust later.

michael:
It isn't a problem to have extra properties set. They'll be stripped automatically if not part of the schema.

src/converter/bib/BibConversion.js
zuphilip: Some publications are part of a container resource, e.g. a chapter is part of a book (i.e. the book title would be the container title) and a journal article is part of a journal (i.e. the journal title would be the container title). In German such publications are also named [unselbstständige Werke](https://de.wikipedia.org/wiki/Unselbst%C3%A4ndige_Literatur). I am not sure whether for the end user something like "container title" is clear and therefore maybe "book title" and "journal name" are better to present in the GUI.

src/converter/bib/BibConversion.js
zuphilip: Okay. Yeah, the ISBN is for example not always suitable as an unique identifier (publisher reuses old ISBN for newer editions etc.).

src/converter/bib/BibConversion.js
michael: Understand. Yeah I guess in future containers and sub publications should even be linked together in a standardised way. However for now I guess we just accept the redundancy and treat them separately. We have a similar situation with affiliations as there often the university is the same but the department is different. However we don't model that hierarchy.

src/converter/bib/BibConversion.js
michael: Created a separate issue, which we can tackle when there's time and we've chosen what strategy we will apply: https://github.com/substance/texture/issues/483

src/converter/bib/BibConversion.js
zuphilip: No problem to have (possibly) some redundancy. And that is actually the same as other bibliographic data formats do like CSL-JSON, RIS, etc. do. Moreover, the possiblity in BibTeX to use links, e.g. [here](http://dblp.uni-trier.de/rec/bibtex/conf/post/NematiBGD18), is IMO rather uncommon practice. And as long as people can rely on some automatic way to receive all the data, they don't care about the details in the background;-)

The same affiliation might be much more likely to occur...

Issue #481, opened 2018-04-16
Real world examples for all supported pub types


michael:
We are currently improving the support for different publication types within the ref list. It would be helpful to have a real world example for each type so we can understand them better and cover edge cases.

@Melissa37 @fabiobatalha @gustavofonseca could you collect and post those here? Maybe including a link to the published article that contains the pub-type.

Melissa37:
_How do we differentiate between books and book chapters?_ 
They are both book-type publications, but a chapter in a book has more metadata (usually the chapter title, page numbers and authors (if the main book is edited and has different contributors for each 
chapter)).

```
<!-- Book reference: Editors, no chapter details -->
    <ref id="bib8">
        <element-citation publication-type="book">
            <person-group person-group-type="editor">
                <name>
                    <surname>Cardé</surname>
                    <given-names>RT</given-names>
                </name>
                <name>
                    <surname>Millar</surname>
                    <given-names>JG</given-names>
                </name>
            </person-group>
            <year iso-8601-date="2004">2004</year>
            <source>Advances in insect chemical ecology</source>
            <publisher-name>Cambridge University Press</publisher-name>
            <publisher-loc>Cambridge</publisher-loc>
        </element-citation>
    </ref>
<!-- Book reference: Has edition, editor is author -->
    <ref id="bib10">
        <element-citation publication-type="book">
            <person-group person-group-type="editor">
                <name>
                    <surname>Chmeil</surname>
                    <given-names>N</given-names>
                </name>
            </person-group>
            <year iso-8601-date="2008">2008</year>
            <edition>2</edition>
            <source>An Introduction to Work and Organizational Psychology: A European
                Perspective</source>
            <publisher-name>Wiley-Blackwell</publisher-name>
        </element-citation>
    </ref>
<!-- Book reference: publisher location - US: city followed by state code -->
    <ref id="bib11">
        <element-citation publication-type="book">
            <person-group person-group-type="author">
                <name>
                    <surname>Coyne</surname>
                    <given-names>JA</given-names>
                </name>
                <name>
                    <surname>Orr</surname>
                    <given-names>HA</given-names>
                </name>
            </person-group>
            <person-group person-group-type="editor">
                <name>
                    <surname>Otte</surname>
                    <given-names>D</given-names>
                </name>
                <name>
                    <surname>Endler</surname>
                    <given-names>JA</given-names>
                </name>
            </person-group>
            <year iso-8601-date="1989">1989</year>
[...]

PR #484, opened 2018-04-19
Improve Bibliography Datamodel

michael:


michael:
@zuphilip please review. I gave you push access to the repo so you can add changes to this branch directly.

I'm happy to normalise our internal model according to the CSLJSON spec where it makes sense. E.g. we could drop the remaining `source` properties and turn them into `containerTitle`. We just need to make sure `EntityConverters.js` and `BibConversion.js` are updated  accordingly.

data/kitchen-sink/manuscript.xml
```diff
@@ -250,19 +250,166 @@
...
+          </person-group>
+          <year iso-8601-date="1959">1959</year>
+          <volume>3: Formulations of the Person and the Social Context</volume>
+          <source>Psychology A Study Of A Science</source>
+          <publisher-name>Mcgraw-Hill Book Company</publisher-name>
+        </element-citation>
+      </ref>
+      <!-- Book Chapter reference: chapter in book with editors -->
+      <ref id="chapter1">
+        <element-citation publication-type="chapter">
```
zuphilip: @michael Any reason you changed the `publication-type` here compared to what is given as a real example `bib48` in https://github.com/substance/texture/issues/481? Currently both should work, because I added the support for the other one in https://github.com/substance/texture/pull/484/commits/49fbf882c8502eadad6ac0f2e89dc2e0417ec1dc.

zuphilip:
There is still some usage of the now deprecated `articleTitle`, `chapterTitle` but I don't know whether this is a problem:
```
$ grep -rl articleTitle
test/converter/ImportEntities._test.js
dist/texture.js.map
dist/texture.js
src/editor/components/TitleGroupComponent.js
src/TextureArchive.js

$ grep -rl chapterTitle
test/converter/ImportEntities._test.js

$ grep -rl dataTitle
```

I have some more ideas to possibly improve the data model, but I would suggest to first finish this PR here, because this is IMO already quite large. Moreover, I guess we also should discuss further with the others.

data/kitchen-sink/manuscript.xml
michael: I thought it may be more explicit to introduce a separate type in JATS as well (less conversion magic). I'd personally be fine with going close to CSL/JSON also on the JATS level, as long as the requirements of the publishers are met.

One question: Does CSL/JSON allow basic formatting (e.g. italics in titles)

zuphilip: Ah, I see. Going close to CSL-JSON does sound fine for me, but I guess it needs some discussions with the others.

> One question: Does CSL/JSON allow basic formatting (e.g. italics in titles)

There are some HTML tags which Zotero imports e.g. in title field and can be used for the citations. A recent commit https://github.com/zotero/translators/blob/d1312c62730a23d936a6862f5e2a289420cdd371/Crossref-REST.js#L47-L55 might show the capabilities and you can see an example [here](https://github.com/zotero/translators/blob/d1312c62730a23d936a6862f5e2a289420cdd371/Nature%20Publishing%20Group.js#L1714 ):
```
"title": "Structure of the SAM-II riboswitch bound to <i>S</i>-adenosylmethionine",
```

However, I don't know whether this is part of the specification or just of the (standard) implementation, but I guess it doesn't change much.

michael: Absolutely. I'm updating the Dar spec accordingly, and then we can have everyone review the fields. Our strategy for Texture is to only add what is commonly agreed on, and leave out edge cases until they are requested, fully understood and confirmed by everyone. I've moved your other suggestions to separate issues.

As for formatting, I'd allow it for certain fields, but go with the semantical HTML tags. E.g. `em`, `strong` instead of `i` and `b`. We could consider `i` and `b` on import but normalise on export. How does that sound to you?

Will move this into a separate issue too.

michael:
`articleTitle` in other places means article title of the full publication, so that is ok. The tests we should bring back at some point. They are currently disabled.

michael:
Merging this. Let's move on with additional PR's. Happy to consider further improvements, generalisations of the data model!

Great work!

data/kitchen-sink/manuscript.xml
zuphilip: As far as I can say that sounds fine. Note that `<sup>` and `<sub>` might be relevant e.g. for chemical formulas. Maybe also MathML for mathematical formulas...

michael:
And then I forgot to actually push the merge button. ;)

Issue #488, opened 2018-04-22
Substance as peerDependency?


oliver----:
We need to investigate how to use `substance` as a dependency correctly.
E.g. Stencila is using Texture as a library, but itself depends on substance.
If both have substance as a regular dependency, it may now happen, that Texture uses a different substance version than Stencila.

Don't know how to do this exactly. My first try would be, adding `substance` to the `devDependencies` and as a `peerDependency`.

oliver----:
@michael what is your take on this?

michael:
Sorry for the delay. I think what you proposed works well; ensures that Stencila does not end up with two different substance dependencies accidentally right? I mean at least you get warned.

oliver----:
No worries. This just happened to me while working on Stencila.
With the current preview versions npm does not give a warning. Hope that this would be different with regular semvers.

oliver----:
We need to see if npm is complaining once we have a regular version string.

PR #489, opened 2018-04-22
Substance as peerDependency.

oliver----:
See #488


oliver----:
Merging this as it works for me and I need to move on in Stencila dev.

PR #493, opened 2018-04-23
Math

Integral:
Why: We want to edit math formulas!
What: We are using JATS inline-formula and tex-math elements as model and KaTeX for formula editing. 
This PR should go after [this substance PR](https://github.com/substance/substance/pull/1175). We still need to improve usability (for instance when you press enter formula component must rerender itself). @michael Please play and let me know what else should be improved.


michael:
Will review today!

michael:
Good as a first implementation. Will merge and create separate issues for things that should be improved.

PR #510, opened 2018-05-03
Collapse selection after ext-link insertion

Integral:
Why: At the moment edit prompt isn't showing after ext-link insertion, so an extra click needed.
What: This PR contains new command and config.

@michael pls review, I had to insert new config (comand, icon, label, keyboard shortcut) in the middle of annotation tools definition. It's not very beautiful, but I can't see other ways to save an old position of the tool. 
I think that the more elegant way could be to add `spec.command` and set `AnnotationCommand` as a default to addAnnotationTool. What do you think? Than we will still be able to call config.addAnnotationTool

### DIFF
--- a/src/editor/EditorPackage.js
+++ b/src/editor/EditorPackage.js
@@ -56,6 +56,7 @@ import TableFigPreview from './components/TableFigPreview'
 
 import DecreaseHeadingLevelCommand from './commands/DecreaseHeadingLevelCommand'
 import IncreaseHeadingLevelCommand from './commands/IncreaseHeadingLevelCommand'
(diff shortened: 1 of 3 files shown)


michael:
Agree with your suggestion. I will take that over so we get this merged faster.

oliver----:
Or is it this way already?

michael:
`addAnnotation` tool is a higher level API of Texture. We could consider moving this into core, but let's evaluate again when we prepare the Substance release?

michael:
It's slightly opinionated (could be app specific), so would personally be fine to leave it as is (so Substance also has the low-level stuff).

PR #511, opened 2018-05-04
Make Dar robust against missing files

michael:
Why: A common source of the error was that StencilaArchive/TextureArchive crashed because of a missing sheet.xml
What: Make sure, that a DAR can still be opened, and a warning is printed.

### DIFF
--- a/src/TextureArchive.js
+++ b/src/TextureArchive.js
@@ -22,6 +22,9 @@ export default class TextureArchive extends PersistedDocumentArchive {
 
     entries.forEach(entry => {
       let record = rawArchive.resources[entry.path]
+      // Note: this happens when a resource is referenced in the manifest
+      // but is not there actually
+      // we skip loading here and will fix the manuscript later on
       if (!record) {
         return
       }
--- a/src/TextureArchive.js
+++ b/src/TextureArchive.js
@@ -21,6 +21,7 @@ export default class TextureArchive extends PersistedDocumentArchive {
     sessions['pub-meta'] = pubMetaSession
 
     entries.forEach(entry => {
+      let record = rawArchive.resources[entry.path]
       // Load any document except pub-meta (which we prepared manually)
       if (entry.type !== 'pub-meta') {
         // Passing down 'sessions' so that we can add to the pub-meta session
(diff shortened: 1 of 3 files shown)


oliver----:
This would not lead to a 'dirty' manifest, would it? Do we want this?

michael:
It removes the missing files as they would cause troubles anyways. If we left those manifest entries there, we would still see the tabs for the missing documents, causing them to break when tried to be opened. As a consequence the archive would be 'cleaned up' automatically. Like when you load and save, the missing entry will be gone from the manifest. Think that is ok, but we should communicate to the user that we cleaned up, or normalised a broken archive.

oliver----:
I mean, would the archive be considered 'dirty'? I.e. saving have an effect?

michael:
Oh good point. That would be the consequence yes. 

So the whole scenario would look like this:

1. Open broken archive
2. Tell the user the archive is broken, do you want to open it anyway? (we can add that ui later, and do it automatically for now)
3. If yes we apply the fixes (which should produce changes to the archive)
4. If the user hits the save button the fixed archive gets stored to disk

oliver----:
That would be desirable. And now my question: does the current implementation already have the effect of marking the DAR as dirty? ;)

michael:
I can perform the manifest fixes on the manifest session instead of on the rawmanifest, so we get that dirty state. That be ok for you?

oliver----:
Maybe that is cleaner... then we have ops too.

michael:
Ok will do.

It's actually good to think about this. We'll have similar situations when loading old archive versions, where we need to clean them up to 'upgrade'. Could be the same mechanism.

oliver----:
Yeah, good point.

michael:
@oliver---- pls review

src/TextureArchive.js
```diff
@@ -22,6 +22,9 @@ export default class TextureArchive extends PersistedDocumentArchive {
 
     entries.forEach(entry => {
       let record = rawArchive.resources[entry.path]
+      if (!record) {
```
oliver----: Maybe leave a comment here, when this happens.

oliver----:
Apart from that, looking good.

oliver----:
Have done it.

oliver----:
@michael BTW, is there a reason to use hierarchical versions like in  '1.0.0-preview.1.8' instead of just increasing the first number?

PR #512, opened 2018-05-04
Clarify roadmap

### LINKED ISSUE #501: Roadmap question: is live collaborative editing planned?
Texture looks amazing! I have not yet set up a test installation, so it's not clear to me: is live, simultaneous collaborative editing by multiple authors (as in applications like Etherpad, Google Docs, and Dropbox Paper) a current feature, or potentially planned? I see Substance supports concurrent collaborative editing, so I hope Texture will as well. Thank you!

brainwane:
Annotate the description of the "group authors" feature to clarify that it also covers live collaborative editing, since that's how some users think of it.

Fixes #501.


michael:
Hi thank you for that PR. However "Group authors" in this case is a different feature. It just means that author groups can be tagged in the document's metdadata. Could you instead add another entry "Realtime Collaborative Editing" without a ready date? :)

brainwane:
Thanks, @michael -- I had been working from https://github.com/substance/texture/issues/501#issuecomment-386652629 . Fixed and ready for re-review.

PR #516, opened 2018-05-07
Lists

oliver----:
# Why?

Too long we have been waiting for this.

# What?

Adds an adapter for Substance List support that works with an XML element.
To have a 'safer' life we should come up with a restricted schema soon.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,10 +9,19 @@ module.exports = {
     "ecmaVersion": 6
   },
   "extends": "eslint:recommended",
-  "globals": {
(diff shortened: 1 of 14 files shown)


oliver----:
@michael @Integral  first but already pretty good support for lists.
I have improved the implementation in substance, and created a first J2T converter.
Pls try out: http://localhost:4000/?archive=list

oliver----:
texture-lists

oliver----:
Consistent nested level types:
texture-lists-nested-list-types

PR #523, opened 2018-05-15
First Iteration on Tables

oliver----:
# Why?

Tables have not been implemented sufficiently.

# What?

This PR introduces a Table implementation based on a very semantic internal model.
Feature-wise this is pretty complete:
- basic interaction
- header cells
- copy'n'paste
- merge/unmerge
- insert rows/cols


oliver----:
texture-table

oliver----:
Deleting a row leads to a bug

michael:
Looks nice with the table selection interaction!

oliver----:
Will add some tests before this PR is ready.

michael:
Works nicely. It's a really minimal solution!

Few suggestions:

- right clicking should first select the cell and then bring up the context menu (currently the selection stays in the previously selected cell)
- if add a column/row we should probably try to replicate the style (e.g. make the first cell of a column a header if the previous one was a header too); potentially also continue with letters / numbers if we can detect that
- Would like to try opening up all the isolated nodes (figures, tables, disp-quote) and adding a handle (e.g. top-left) to select the full node.

michael:
Should we get this merged and make a Texture Preview 2 release? We could trigger the consortium members for testing the new features (lists and tables).

oliver----:
@michael this 'UI' testing is pretty cool :)

oliver----:
I think we are good for now, and can merge.
I have established a foundation for testing the UI. This needs to be completed at some point, but good enough for now.

oliver----:
@michael 
Your first suggestion was just a bug, which should be fixed now.
The second is on my list.
The last needs some collab...

michael:
Yay! Of course I'm in for the UI improvements. But this is already pretty good, so not blocking a release I would say! :)

PR #545, opened 2018-05-28
Reader (Iteration 1)

michael:


src/editor/components/TextureContainerEditor.js
```diff
@@ -8,9 +8,10 @@ export default class TextureContainerEditor extends ContainerEditor {
 
   _renderNode($$, node) {
     let api = this.context.api
+    let disabled = this.props.disabled
     if (!node) throw new Error("'node' is mandatory")
 
-    let props = { node }
+    let props = { node, disabled }
```
oliver----: I will fix this in substance land

oliver----: If you wait some minutes I can push a better version.

michael: please push :)

oliver----:
Yeah.. I am about to do that. Just our yesterdays work needs some 'polishing' ;)

src/editor/components/TextureContainerEditor.js
oliver----: Yeah.. I am about to do that. Just our yesterdays work needs some 'polishing' ;)

michael: I'm working on #564 right now... Take your time.

oliver----:
@michael we need to talk about how we want to proceed with the 'master'/'develop' pattern. We had and are still having a situation now, where we want to continue development while other projects a release, which need some patches over time.
We could either use 'master'/'develop', or 'release'/'master'. But we need two tracks.

oliver----:
And then I'd suggest to use the same pattern everywhere.

michael:
@oliver---- yes I understand. I'd prefer for release/master. And master being the default branch, so contributions always target the latest development version.

oliver----:
Ok. I have some things on 'develop' which I will merge in master.
Create a release branch on the commit we used for the last release.
And this branch need to be rebased.
Are you having uncommited changes?

michael:
I have some very experimental stuff, that  I don't want to commit yet. But please go ahead and i will use git stash to bring back my changes.

oliver----:
ok.

michael:
Can you fix the editor so it also provides the api in context? It's currentry broken.

oliver----:
Will do after the rebase.

michael:
@oliver---- I think we also need to make the referenceManager available to the API. I guess then the component implementations could talk to the ReferencesModel which abstracts all that label generation etc. away. 

What do you think?

First thing I could do is creating the instances earlier and passing them to the API. Next we should fix places that use referenceManager etc. directly and instead use the new ReferenceModel API.

oliver----:
Yep. Everything need to go into the API. Like it was a unified model. Of course with all the 'managers' inside to do the work.

michael:
Ok i will try to start this work then. You can forget about the broken editor, I will fix as part of this restructuring.

oliver----:
Ok. But let me finish the rebase. There are some regressions I need to fix.

oliver----:
@michael Now you can start.

oliver----:
I need to fix test coverage reports in the meantime. (using the same toolchain as in stencila/engine et al.)

michael:
Okay. I only have one hour left today, so probably won't finish this today.

oliver----:
Ok

oliver----:
But maybe you could push your stuff, even unfinished. Otherwise I am a bit blocked tonight.

michael:
Sorry saw that too late. But you can move on and push, i will rebase my stuff and also push asap

michael:
@oliver---- I just pushed that change that moves managers into the API. 

It's basically working but somehow the setup of managers within commands seems not right. When you open the editor, and set the cursor to the title, I get this error:

image

It's a lifecycle thing, like on construction of the command manager we don't yet have setup the managers and API. Don't know how to approach this. I think we generally want to get rid of managers inside editorSession... just don't want to kill too much at once... as it's affecting a lot of code.

Could you take a look?

oliver----:
Will take a look.
I will rebase on the latest master.

oliver----:
@michael this is working now. Pls take a look, particularly at https://github.com/substance/texture/pull/545/commits/0e838c16a3c2983b81f275da47e9f2706fab4e88

michael:
Thank you @oliver---- . Makes sense to do the monkey patch for now. And in the long run let's get all this stuff out of editorSession.

oliver----:
Calling it a first iteration.

PR #565, opened 2018-06-06
Extracted tableHelpers.

oliver----:
# Why?
Substance provides now some general helpers for tables, e.g., how to create column labels 'ABC' from index, etc.

# What?

This PR removes these helpers from this repo.


oliver----:
Merging this because the 'failing' is because istanbul does not support es8 yet.

PR #576, opened 2018-06-21
Figures preview

Integral:
Figures and table preview for the reader.

### DIFF
--- a/src/article/models/FigureModel.js
+++ b/src/article/models/FigureModel.js
@@ -50,4 +50,5 @@ export default class FigureModel extends DefaultModel {
     const content = this._node.findChild(contentType)
     return content
   }
+
 }
--- a/src/reader/ReaderPackage.js
+++ b/src/reader/ReaderPackage.js
@@ -16,7 +16,6 @@ import UnsupportedInlineNodeComponent from '../editor/components/UnsupportedInli
 import BreakComponent from '../editor/components/BreakComponent'
 import ContainerNodeComponent from '../editor/components/ContainerNodeComponent'
 import ElementNodeComponent from '../editor/components/ElementNodeComponent'
-import FigComponent from '../editor/components/FigComponent'
 import CaptionComponent from '../editor/components/CaptionComponent'
 import FrontComponent from '../editor/components/FrontComponent'
 import GraphicComponent from '../editor/components/GraphicComponent'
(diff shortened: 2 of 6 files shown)


Integral:
@michael i'm not sure that it was a good idea to render table and figures inside reader xref component. On the one hand i think it was a good idea to keep html rendering inside model, on the other hand rendering tables without component could be a bit tricky and messy. What do you think, should i move it to models?

michael:
I chose to render inside xref component because of positioning (may also change in future but was quickest for now). 

However as you say we could move the rendering code into separate components. The ideal goal would be to use `shared/FigPreviewComponent` etc. that is then used in the reader but also in the writer when selecting the targets.

Btw, @oliver---- started a big code restructuring of Texture, so it's likely we produce conflicts with new changes while he's working on that.

@oliver---- is it ok to continue this work, or better wait for your first round to be finished?

michael:
@Integral in the meanwhile can you try out if it feels better if preview popups are shown on explicit click of xref, compared to mouseover as done currently? Also we need to jump to functionality, when you click on the fig preview / table preview etc element, it should take you to the object on the page.

Integral:
Let use anchors for jumping? Do we need highlighting?
Regarding popup on click, how can we jump in this case? Do we need to
render button inside popup then? For jumping I mean.

Integral:
@michael do we want to render [1,2] as to separated links inside brackets?

Integral:
@michael i've tried to do it [here](577) very roughly. We should decide something here as making a links inside labels [1,2], figure 1,2 or [1 to 3] are very tricky...

Integral:
Yes!

Integral:
@michael I did try this in #579. However there are couple of problems as well. First is that there must be a link inside bib ref. So I'm not sure that we can wrap anything outside the label into the link. The other thing that we need to detect clicks/escape key to close popup and we need kind of the trap thing for it.
Anyways check if you like it and I will continue.

michael:
Forget about closing the popups for now. We can fix that later. But we need bigger click targets. Like in the popup the whole figure preview should be clickable same for references. Maybe add some hover style. Just try to make it look nice. And for xref, we should render them as `<a>` tags or `<button>`> not as span.. that makes for a strange click behavior.

Please finish that and let's merge it. Then you can do footnotes.

Merci. :)

michael:
Looks good. Merging this iteration. Maybe take care of footnotes display next. Using the API.

PR #594, opened 2018-07-07
Towards a new metadata editor.

michael:


michael:
@Integral I'd be for using a custom Select implementation. Then we can use the overlay logic in the same way. And it also looks more minimal.

oliver----:
This has become a longer journey with some side tracks. Feeling confident with the new concepts, we continue development on master.

PR #604, opened 2018-07-14
App initialization logic rewritten

meinbrodt:
Hi Everybody, 

the whole rewrite tries to achieve the following things: 

- Enforce better separation of concerns. For instance, the `texture/src/article/ArticleLoader.js` class loads the article document but creates also the texture editor session. This two steps have been separated into two service classes: (1) `texture/src/sessions/EditorSessionsGenerator.js` and (2) `texture/src/dar/DocumentLoader.js`. Furthermore, I "moved" a lot of functions from the `PersistedDocumentArchive` and `TextureArchive` class to dedicated external service classes.

- A cleaner and more easy to use approach (in my opinion) to configure an application or better it's contained parts (dar and dar-clients) using dedicated configuration classes instead of using getter methods on the app classes.

- A separation of the those parts of the code necessary for "read-only apps" and "read-write editor apps". The central change here is the replacement of the `texture/src/dar/PersistedDocumentArchive.js` class with two new classes: 
(1) `texture/src/dar/DocumentArchiveReadOnly.js` (contains only those methods necessary for displaying a dar) 
(2) `texture/src/dar/DocumentArchiveReadWrite.js` (inherits all methods from the read-only class and furthermore contains all the necessary methods for editing and saving a dar). 
However, also several configuration and service classes follow that logic (e.g. see the `texture/src/dar/DocumentArchiveReadWriteExporter.js` and `texture/src/dar/DocumentArchiveReadOnlyExporter.js`)

- Achieve a better understandability of the code via promises and promises chaining. A good example are the load methods of the `DocumentArchiveReadOnly` and `DocumentArchiveReadWrite` classes.

I know there are a lot of changes and I don't expect that all or any of them to be accepted, however I hope that this can start a conversation on how to further improve the code.

Last but not least, I just wanted to say that I really like the project and hope to use it for some of our projects. If you need any help with something just let me know. 

Regards, Matthias

PR #616, opened 2018-07-18
Rewrite app init code after qc

oliver----:
Taking over PR #604 for discussions.


oliver----:
Hey @meinbrodt 
Using this PR to discuss your original PR #604 and moving further.

There is a bit of a concept drift currently going on. I.e. we are considering a DAR agnostic of specific resource types. That means, our article should just be one option. This we want to address by having a configurator with nested configurations for the specific resource type.
We consider Texture now an editor with tabs for multiple resources.
Along with the article you could have a csv file in the DAR that is edited via a Spreadsheet interface.
So our original approach having an 'Substance.EditorSession' for each resource is obsolete.

oliver----:
@meinbrodt have sent you an invite for collab

meinbrodt:
> ... we are considering a DAR agnostic of specific resource types. That means, our article should just be one option. 

Great idea! I suppose the correct load, transformation and import routines (e.g. which importer, resource configurator etc. to use) would then be kicked off based upon in information about the resource contained in the manifest.xml file of the DAR right?

I think - correct me I'm wrong - this idea is already reflected rudimentarily in my implementation of the dar/DocumentLoader (see here: https://github.com/substance/texture/blob/rewrite-app-init-code-after-qc/src/dar/DocumentLoader.js) 

> We consider Texture now an editor with tabs for multiple resources.

Does that mean, that it won't be possible anymore to create a reader app without editing capabilities?

> So our original approach having a 'Substance.EditorSession' for each resource is obsolete.

This I don't fully understand, however, I still need to grasp the concept of a session. I haven't advanced this far yet.

oliver----:
>> We consider Texture now an editor with tabs for multiple resources.
> Does that mean, that it won't be possible anymore to create a reader app without editing capabilities?

At least in the default configuration when run as app. A read-only version will be added, too, once as a preview in the editor (which could be used as a standalone reader, too) and as a command-line site-generator.

oliver----:
>> So our original approach having a 'Substance.EditorSession' for each resource is obsolete.

> This I don't fully understand, however, I still need to grasp the concept of a session. I haven't advanced this far yet.

This just makes the DAR loading a bit simpler.

oliver----:
Closing this, because it is outdated. We will rework the whole App initialisation code soon.

PR #622, opened 2018-07-20
Dedicated editor for translations.

michael:
@oliver---- can you check if this is going into an ok direction?

@Integral there's more to do here and I could need your help, but just finish other task first.

### DIFF
--- a/data/kitchen-sink/manuscript.xml
+++ b/data/kitchen-sink/manuscript.xml
@@ -12,6 +12,9 @@
       </article-categories>
       <title-group id="title-group-1">
         <article-title id="article-title-1">Object vision to hand action in macaque parietal, premotor, and motor cortices</article-title>
+				<trans-title-group xml:lang="es">
+					<trans-title>Si <italic>signore</italic></trans-title>
+				</trans-title-group>
       </title-group>
       <contrib-group id="contrib-group-1" content-type="author">
         <contrib id="contrib-1" contrib-type="person" equal-contrib="yes" corresp="no" deceased="no">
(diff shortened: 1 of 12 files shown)


oliver----:
@michael hard to say atm. I guess the challenge here is to find a Model for translations, that can be implemented well on top of JATS, but is general enough. I'd say, go ahead with your approach, and then let's see ;).
Maybe, at some point you could capture a version of a model for translations, without having the JATS based impl already in mind. (Or is it that already?)

michael:
that's what I'm trying to acchieve with that Translatable API. I just define translatables, which have an id. E.g. `article-title` or `abstract` and then we can modify the original language as well as adding and updateing translations.

E.g. `model.addTranslation('es')`

`model.getTranslations()` will yield translation models.

oliver----:
Sounds good. This sounds a bit like a mixin.. But for now we will only have titles and abstracts...

michael:
We could also use it for figures then it gets a bit more of a dynamic nature, that's why I'm using ids for the translatables. E.g. each figure could produce a virtual view on the caption translatable.

`api.getTranslatable('fig1-caption-trans')`

The idea would be that we have all translations in one place (task-based for the translator person). So I would maybe not show them in the Figures section.  (if that's what we want)

oliver----:
Yep. That is something we need to discuss with Editors.
- Dedicated Translations section
- Adding to cards in other collections (e.g. Figures)
- Or both

michael:
Let's do a first shot, and present it to them. For the metadata editor demo i will leave it for title + abstract for now.

Issue #614, opened 2018-07-18
Import and display article record (including pub history) in MetadataEditor


michael:
This input is already in the kitchen sink, but we need to make it editable.

```xml
      <pub-date id="date-5" publication-format="print" date-type="pub" iso-8601-date="1999-01-29">
        <day id="day-5">29</day>
        <month id="month-5">01</month>
        <year id="year-5">1999</year>
      </pub-date>
      ...
      <history id="history-1">
        <date id="date-1" date-type="received" iso-8601-date="1998-01-05">
          <day id="day-1">05</day>
          <month id="month-1">01</month>
          <year id="year-1">1998</year>
        </date>
        <date id="date-2" date-type="rev-request" iso-8601-date="1998-03-14">
          <day id="day-2">14</day>
          <month id="month-2">03</month>
          <year id="year-2">1998</year>
        </date>
        <date id="date-3" date-type="rev-received" iso-8601-date="1998-05-24">
          <day id="day-3">24</day>
          <month id="month-3">05</month>
          <year id="year-3">1998</year>
        </date>
        <date id="date-4" date-type="accepted" iso-8601-date="1998-06-06">
          <day id="day-4">06</day>
          <month id="month-4">06</month>
          <year id="year-4">1998</year>
        </date>
      </history>
```

Plan is the following: Introduce an ArticleRecord model (EntityDatabase.js), it's a singleton with a fixed id: "article-record".

```js
export class ArticleRecord extends DocumentNode {}

ArticleRecord.schema = {
  type: 'article-record',
  volume: { type: 'string', optional: false, default: '' },
  issue: { type: 'string', optional: false, default: '' },
  fpage: { type: 'string', optional: false, default: '' },
  lpage: { type: 'string', optional: false, default: '' },
  pageRange: { type: 'string', optional: false, default: '' },
  elocationId: { type: 'string', optional: false, default: '' },
  acceptedDate: { type: 'string', optional: false, default: '' },
  publishedDate: { type: 'string', optional: false, default: '' },
  receivedDate: { type: 'string', optional: false, default: '' },
  revReceivedDate: { type: 'string', optional: false },
  revRequestedDate: { type: 'string', optional: false }
}
```

So this needs to be converted on import and export. Please make sure the following criteria is fulfilled:

- [ ] we use fixed date-types (pub, received,accepted, rev-received, rev-requested)
- [x] pub-date element is used for storing publishedDate (directly inside article-meta element)
- [x] other date properties are stored inside `history`
[...]

PR #625, opened 2018-07-22
WIP: Import and display article record.

Integral:
Keep issue, volume, fpage, lpage, page-range, elocation-id, pub-date and history inside a singleton entity. Detailed description given in issue #614


Integral:
@michael there is no visual editor yet, but could you please review other things?

michael:
@Integral I think it's better to drop all fpage,lpage stuff from the XML in InternalArticle, and use pubmetaDb node instead. So converter should take Source XML (TextureArticle.rng) and convert to ArticleRecord entity. 

https://github.com/substance/texture/blob/metadata-editor/src/article/metadata/EntityDatabase.js#L5-L18

We can immediately create an article-record node, with `id: 'article-record'` and then fill it up during the conversion process. We could also put author array there.

PR #632, opened 2018-07-24
Workflow et al.

oliver----:
- enabling mutually exclusive overlays
- first workflow


oliver----:
@michael #wasdaseh

exture-exlusive-overlays

Issue #609, opened 2018-07-17
Introduce Toolbar for MetadataEditor


michael:
We want to provide a dedicated toolbar for the metadata editor, to allow quick access to related tasks (e.g. adding entities such as authors, affiliations etc.)

The toolbar will include a mode switcher (right side) to be able to switch between editor, metadata editor and reader (=preview) modes.

![image](https://user-images.githubusercontent.com/2931/42836385-300c9f1e-89fb-11e8-9cb9-8d1f6a2bcd68.png)

michael:
Also see #610 for more UI/UX ideas including batch actions.

michael:
@Integral for the coming days.. could you do a first iteration on this:

- [ ] Create a tool for the mode switcher (dropdown). At first let's just use the app toolbar as it is, but use fixed positioning to place it over the editor toolbar in the right upper corner (that's fine for a start i think)
- [ ] Add tools in Metadata editor for adding a new Author, Editor, Keyword, Award, Reference, Group etc.

oliver----:
We have this now.

PR #633, opened 2018-07-30
Metadata toolbar

Integral:
Toolbar improvements for metadata editor. See #609


michael:
It's going into the right direction. I'm merging this! :) Thank you!

PR #640, opened 2018-08-01
Improved model for references

michael:
This PR removes the (redundant) XML based model for references from the internal article model,
and adapts the other implementation accordingly.
There are regressions introduced by this PR.


oliver----:
We will fix the regressions later on.

PR #653, opened 2018-08-10
Refactor internal schema

oliver----:
# Why?

With introducing the Model API we now feel confident to improve the InternalArticle schema.
Up to now, the model was pretty kept close to JATS. However lately we started to deviate more and more. Now the difference is already too large to accept the drawbacks of using that model.

# What?

Instead of working on a quasi JATS model this PR introduces a hybrid Document schema, that uses custom nodes for metadata, and JATS elements for content.


oliver----:
We will merge this knowing that this introduces a lot of regression.
In the next iteration we will make use of the simpler model, particularly leveraging the explicit nodes for so called 'collections'.
In another branch we will work finishing the importer and getting the exporter running again.

PR #658, opened 2018-08-15
Importer/Exporter rewrite

oliver----:
# Why?
We are moving towards a more customized internal schema, which is easier to work with.
A key stumble stone has been the DOM transformation based importer/exporter.

# What?

This PR introduces a rewrite of the converter, that generates from JATS DOM into our InternalArticleSchema, which is a hybrid XML schema with custom extensions.


oliver----:
On a good way.

PR #660, opened 2018-08-17
Language editor

Integral:
This PR introduces a language editor for things like keywords and subject where you can pick a language.


michael:
Looking good. Merging this.

PR #661, opened 2018-08-20
Toggle xref targets via models.

Integral:
With switching to the model-driven approach we want to ask models to change xref targets and perform transactions in the Article API.


Integral:
@oliver---- this is working, however there are some rerendering glitches (like labels didn't updated so fast), could you look at this thing?

PR #663, opened 2018-08-20
Article Reader

oliver----:
- Introducing another article tab that shows a read-only version of the article
- Tidy-up the editor implementation to increase code share
- Remove dead code


oliver----:
Good for a first iteration.

oliver----:
In a next round, the TableComponent must be changed so that it obeys to 'context.editable'.

PR #664, opened 2018-08-22
Overhaul xref editing

michael:


oliver----:
@Integral I have rebased your branch onto latest master. Of course there were a lot of conflicts which I tried to resolve.
Maybe you could check again if this is now as you expect.

PR #676, opened 2018-08-24
Repair add reference workflow

Integral:
Another iteration on repairing add reference workflow.

### DIFF
--- a/src/article/models/ReferenceCollectionModel.js
+++ b/src/article/models/ReferenceCollectionModel.js
@@ -20,9 +20,4 @@ export default class ReferenceCollectionModel extends DefaultCollectionModel {
   removeItem (item) {
     this._api.deleteReference(item, this)
   }
-
-  _prepareItem (item) {
-    item.type = this._getItemType()
-    return item
-  }
 }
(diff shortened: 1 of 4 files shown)


src/article/models/ReferenceCollectionModel.js
```diff
@@ -9,12 +9,20 @@ export default class ReferenceCollectionModel extends DefaultCollectionModel {
     return result
   }
 
-  // TODO: explain
+  addItem (item = {}) {
```
oliver----: Why do you need to override the parent impl here?

src/article/models/ReferenceCollectionModel.js
```diff
@@ -9,12 +9,20 @@ export default class ReferenceCollectionModel extends DefaultCollectionModel {
...
+
   addItems (items) {
     return this._api.addReferences(items, this)
   }
 
   removeItem (item) {
     this._api.deleteReference(item, this)
   }
+
+  _prepareItem (item) {
```
oliver----: Don't Copy'n'Paste without thinking ;)

oliver----:
As a general request, try to be aware what code is necessary and which could be spared. Otherwise we create dead code from the beginning.

oliver----:
@Integral Pls rebase and merge

PR #681, opened 2018-08-24
Add reference test

oliver----:
- render command names as classes for tools and menu items as a preparation for UI tests
- test opening and closing the 'Add Reference' workflow
- test creating each ref type

### DIFF
--- a/src/article/shared/AddReferenceWorkflow.js
+++ b/src/article/shared/AddReferenceWorkflow.js
@@ -35,7 +35,7 @@ export default class AddReferenceWorkflow extends Component {
     const refTypesButtons = $$('ul').addClass('se-reftypes-list')
     targetTypes.forEach(item => {
       refTypesButtons.append(
-        $$('li').addClass('se-type').append(
+        $$('li').addClass('se-type sm-' + item).append(
           labelProvider.getLabel(item)
         ).on('click', this._onAdd.bind(this, item))
       )
--- a/README.md
+++ b/README.md
@@ -91,6 +91,28 @@ To package the application for distribution do the following:
 $ npm run release
 ```
 
+## Running tests
+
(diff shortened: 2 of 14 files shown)


oliver----:
@Integral ready for you to take over. In some cases we might want to add CSS selectors in the components so that it is easier or cleaner to access buttons etc.

Integral:
@oliver---- could you please help me with configuring async tests map?
[Here](https://github.com/substance/texture/pull/681/commits/45ced556751bc190ba860b258b7d563a7a32b6f2#diff-a0d7fc3a13b214fa7c0fa690995b8eb3R24) I'm trying to run a small test suite against each type of reference, I'm pretty sure we need this because in the past we have some of them broken.
So I'm trying to map those tests for each ref type and run. This is even working, but in a very strange way: it's working slowly and I have a felling that it resolves after some timeout, not with the test ending... Could you look into it?

src/article/shared/AddReferenceWorkflow.js
```diff
@@ -35,7 +35,7 @@ export default class AddReferenceWorkflow extends Component {
     const refTypesButtons = $$('ul').addClass('se-reftypes-list')
     targetTypes.forEach(item => {
       refTypesButtons.append(
-        $$('li').addClass('se-type').append(
+        $$('li').addClass('se-type sm-' + item).append(
```
oliver----: Right. These things we need for UI tests.

oliver----:
@Integral This was looking pretty good already 👍 
I think the only 'missing' piece was an 'await' for the call of the actually async test impl.

oliver----:
Feel free to merge!

oliver----:
There seems to be an issue with MemoryDOM. Need to fix that before we can merge.

oliver----:
@Integral merging this

PR #701, opened 2018-08-29
Select box for Licenses

Integral:
We want to show select box for picking up license from constant values.


oliver----:
Merging this without trying. Code is looking good.

michael:
Maybe we can generalise an EnumEditor and use it in different places (language selection, license selection,...)

michael:
But let‘s do this on the next occasion.

PR #705, opened 2018-08-30
Misc fixes

michael:


Integral:
@oliver---- can you check why node tests are failing (add reference again)? Browser tests are ok, but add reference are working very slooowly again...

oliver----:
@Integral It doesn't look like the tests are broken. The 'Silence' commit was running on Travis successfully. Only it seems that the github hook choked for whatever reason, and the result did not come back. So, I assume you are talking about your local setup.
But @michael you have broken the persistence test with the figure permission converter. I would like to have the test suite green if there is not a good reason to break it.

Integral:
@oliver---- For me a master branch and this branch fails locally when I do ```npm test```

oliver----:
@Integral then you need to fix your local installation, i.e. `npm i`. If that isn't working then wipe `node_modules`. npm is sometimes not updating `node_modules` properly - don't know why.

But in general, if your local version is not working, try to `npm i` and after that `rm -rf node_modules && npm i `.

Integral:
@oliver---- I did try npm i of course. But rm -rf node_modules did the trick now. Thank you.

Integral:
@oliver---- anyways, I was sure that it is still broken, so that's why I asked Michael to merge. Sorry about that.

oliver----:
No worries.

oliver----:
@Integral re performance: the source of the problem is that Metadata- and ManuscriptEditor are doing a full rerender, which seems to take a longer time. To improve that we will need to use refs and override automatic rendering at some places via `shouldRerender()`. But this is not critical, so let's wait a bit.

Issue #672, opened 2018-08-24
Add keyboard shortcuts for metadata editor


michael:
Add author: CommandOrControl+alt+a
Add editor: CommandOrControl+alt+a
Add group: CommandOrControl+alt+g
Add affiliation: CommandOrControl+alt+f
Add award: CommandOrControl+alt+w
Add keyword: CommandOrControl+alt+k
Add subject: CommandOrControl+alt+s
Add footnote: CommandOrControl+alt+o
Add reference: CommandOrControl+alt+r

Integral:
@michael do you know that currently it works only when you put a focus on some input?

Integral:
@michael unfortunately there are more problems
some of combos are not working in chrome, for instance if you do CommandOrControl+alt+w it will do Cmd+W, e.g. it will close tab. The same thing happens with some other combos. Any ideas?

Integral:
I found didn't find enough free combos in CommandOrControl+Alt space... So in #710 we have this:
Add author: CommandOrControl+Shift+A
Add editor: CommandOrControl+Shift+E
Add group: CommandOrControl+Shift+P
Add organisation: CommandOrControl+Shift+O
Add award: CommandOrControl+Shift+Y
Add keyword: CommandOrControl+Shift+K
Add subject: CommandOrControl+Shift+S
Add footnote: CommandOrControl+Shift+F
Add reference: CommandOrControl+Shift+L

But I think that we should need to register global key handler as you said and reject any native behaviour and go back to CommandOrControl+Alt space...

oliver----:
>Let's just use other combos that don't close the window

I guess this is not a good decision.
We are not in an urgent need to get this 'nailed' in a hurry. Instead we should invest our time into work that brings us forward.

I'd say this is blocked until we have the global event handling.

Integral:
Agree. Please keep in mind that besides of browser there are browser extensions which could also use ANY key combos ;)

oliver----:
Ok, gotcha. I guess if we stop up-bubbling keydown events, the browser and plugins should not receive anything.

oliver----:
Let's try.

oliver----:
@Integral do we have an issue for this global event handling already?

oliver----:
I don't think that all originally suggested shortcuts are good: Cmd+Alt+F is for search, Cmd+Alt+S is probably better related to saving...

oliver----:
Beyond that the obstacles are fixed via #783

PR #710, opened 2018-08-30
Metadata shortcuts

Integral:
Keyboard shortcuts for adding new entities via metadata panel.

### DIFF
--- a/src/article/metadata/MetadataPackage.js
+++ b/src/article/metadata/MetadataPackage.js
@@ -102,49 +102,49 @@ export default {
       collection: 'authors',
       commandGroup: 'add-entity'
     })
-    config.addKeyboardShortcut('CommandOrControl+Alt+A', { command: 'add-author' })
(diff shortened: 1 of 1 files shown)


oliver----:
@Integral Which state is this in? I agree with your opinion, that we should rather fix the event handling problem than avoiding certain combos.

Integral:
@oliver---- [here](https://github.com/substance/texture/issues/672#issuecomment-417360274) I reported the current state

oliver----:
Ok. I have read this. Your comment indicates that this PR would be obsolete if we solved the 'global' key handling problem. Is that true?

Integral:
@oliver----, I think we should reassign shortcuts in case of solving global problem. Let's maybe continue here?

oliver----:
Yeah, I think it does not help us if we do this work 10 times. Let's first prepare the work by dealing with the global event problem.

Issue #709, opened 2018-08-30
Deletion of figure breaks


michael:
Known issue, just to track it.

![image](https://user-images.githubusercontent.com/2931/44861421-ca9add80-ac78-11e8-9e03-24517d868208.png)

oliver----:
Thx. Was not aware of it yet :)

oliver----:
Fixed.

PR #713, opened 2018-08-31
Fix Figure Deletion 

oliver----:
## Why

see #709 

## What

This brings back legacy support in the new EditorSession implementation, necessary for some implementation in Substance core to work correctly.


oliver----:
I have checked that this change is not the reason for the red test.

PR #715, opened 2018-08-31
Converter for figure permissions.

michael:


michael:
@oliver---- can you check that vfsSaveHook. I didn't get it to work. Also see #720

oliver----:
THX!

PR #716, opened 2018-08-31
Find and replace (Rewrite)

oliver----:
## Why

Find-n-Replace is broken since we introduced the new AppState API moving away from the former EditorSession oriented implementation. Moreover there have been known performance issues and missing features.

## What

This PR is a refactor / rewrite of the original implementation, with missing features implemented.
Plus: some general tidy-up.

> Note: this is only the first iteration to be continued.


oliver----:
Merging this as a first iteration.

Issue #526, opened 2018-05-18
Tables: replicate cell type when inserting rows


oliver----:
@michael suggested:

if add a column/row we should probably try to replicate the style (e.g. make the first cell of a column a header if the previous one was a header too)

oliver----:
> potentially also continue with letters / numbers if we can detect that
This however does not make sense, because the table is not a sheet. The users will add their headings.
A numbering of 'A - Z' is just arbitrary content, that is only there as a template after creation.

michael:
Oh you are of course right. :)

@Integral you want to take this one?

Integral:
Yes, is do!

Integral:
@michael then here it is #722

oliver----:
> Note: earlier conversation which turned out to lead into the wrong direction.

PR #722, opened 2018-09-04
WIP: Replicate cell type when inserting a new column and row

Integral:
When a new column or row added we want to copy some of the cell attributes.
Currently we want to copy only cell heading attribute, but in future we may want to copy cell types as well.
This PR contains changes to TableEditing interface, we are keeping cell attributes from other column, but currently replicate only heading attribute.

### DIFF
--- a/src/article/shared/TableEditing.js
+++ b/src/article/shared/TableEditing.js
@@ -325,9 +325,10 @@ export default class TableEditing {
         let cellBefore = row.getChildAt(colIdx - 1)
         protoAttributes = cellBefore.getAttributes()
       }
+      let isHeading = protoAttributes.heading
       for (let j = 0; j < n; j++) {
         let cell = $$('table-cell')
-        cell.attr(protoAttributes)
+        if (isHeading) cell.attr({heading: true})
         row.insertBefore(cell, cellAfter)
       }
     }
--- a/src/article/shared/TableEditing.js
+++ b/src/article/shared/TableEditing.js
@@ -318,8 +318,16 @@ export default class TableEditing {
     while (rowIt.hasNext()) {
       let row = rowIt.next()
       let cellAfter = row.getChildAt(colIdx)
+      let protoAttributes
+      if (cellAfter) {
+        protoAttributes = cellAfter.getAttributes()
+      } else {
+        let cellBefore = row.getChildAt(colIdx - 1)
+        protoAttributes = cellBefore.getAttributes()
+      }
       for (let j = 0; j < n; j++) {
         let cell = $$('table-cell')
+        cell.attr(protoAttributes)
         row.insertBefore(cell, cellAfter)
       }
     }


michael:
Good start. However check the following cases:

- only one column is marked as a header (then select this cell and and do insert 1 column right)
- this should also work for rows (e.g. if selected cells of column 1 are headers, then insert headers too for that column

I think we need to define the exact behaviour first. Let me take some time to formalise this.

michael:
consider adding a row when this is selected:

image

then we would expect that:

image

well and if we want to insert 4 new cells based on this selection, we'd want to have alternating headings.

image

To do that we need to define a property, which we call `typeForNewCell(rowIndex, colIndex)` and it returns the cell type that should be used for this new cell based on the current selection. Then this should not just work for the first rows / columns in a table but for everything. You could do some nice painting with this. :) 

@oliver---- what do you think?

oliver----:
Hmm. I think this should be simple. The ability to select a range of cells seems to confuse you.
In GSheets  to insert multiple rows you would select whole rows, not just a sub-rectangle.
We have just not followed this approach, to avoid extra table chrome.
The answer is simply, the attributes of the reference rows/cols should be used.
Like you would copy and insert, with the only difference, that it copies whole rows/cols, not just the selected range.

oliver----:
Note, that this is not only a matter of 'th' only, also probably it should also be applied to col-/row-span.

oliver----:
Looking at GSheets, a maybe better way could be to leave this untouched, but improve pasting.

michael:
In GSheet it also works with regular selections in addition to the explicit column/row selectors:

Consider this: 
image

Adding two cols to the right it takes the format from the most right cell.

image

If you add to the left (here you need to use explicit column selection) it will expand the most-left cell's format.

image

And for rows expansion it works the same:


image

->

image

oliver----:
Ah, ok.

LibreOffice does it without formatting. How is it in Pages?

I assume, Copy'n'Pasting is probably what the people do now, because there you get both, formatting and merged cells. If you think about template then this is closer.
So I am still for improving paste instead of making this fancy.

oliver----:
Just looked at Pages, they seem to do the same as GSheet. I.e. the first (=top-/left-most) row / column is the template for the inserted rows / cols.
Interestingly they do not allow arbitrary places for `th`. Instead one can only have a certain number of header and footer rows and a certain number of header columns.

oliver----:
@Integral after the investigation, pls change the implementation to use the top-most selected row as a template for inserting rows, and the left-most selected column as a template for inserting columns.
This does not include spans.

oliver----:
The much more useful feature will be #725

michael:
Agree to the proposed approach! We can improve later, it's not critical.

oliver----:
@Integral I would be in to drop this and wait until we really need this. The copy and paste behavior I have implemented, and it is so much more useful than this.

oliver----:
I have changed issue #526 according to our findings here.
Closing this PR because we won't fix it now, and not in this way.

Issue #711, opened 2018-08-31
Bring back schema compliant command states and Article manipulation


oliver----:
Since we have switched to the custom internal model, we lost schema-aware commands.
We want to solve this in a general way.
For containers we will use `nodeSchema.targetTypes` which is a list of allowed node-types.
We have not considered 'targetTypes' for text properties, but we could use it for the same purpose.

oliver----:
Done via #738

PR #738, opened 2018-09-06
Schema compliant content creation.

oliver----:
## Why

see #711 

## What

This introduces a CommandManager that takes the current selection and the node schema into account to disable commands that create content.


oliver----:
@michael this PR contains a first round of changes to support rich text fields in references. Please take a closer look, as I suppose you will be the one who is continuing this work.

PR #746, opened 2018-09-08
Card labels

Integral:
## Why

We want to try out rendering of a human readable type of a card.

## What

In this PR every card expecting a label which is rendered above the border as suggested.


michael:
Having the type "person" is fine I think. Just makes obvious that authors and editors have the same target type.

PR #760, opened 2018-09-10
Card tools.

Integral:
## Why

We want to try out a special set of tools for changing card orders and removing them.

## What

In this PR we are introducing a set of commands which are trying to map the current selection to appropriate collection and execute the command where it is possible.


oliver----:
@Integral pls rebase this one. Unfortunately there is a conflict 😸

Issue #758, opened 2018-09-09
Adding of new table is broken


michael:
Repair and add test like we have for creating a new figure.

Integral:
Here it is #773, I think we need a separate issue for tests.

Integral:
This is merged, but hopefully will changed with a JATS snippets.

PR #773, opened 2018-09-11
Insert table

Integral:
## Why

Table insertion is not working as described in #758.

## What

This PR contains a transaction for creation a table-figure element with all nested properties.


Integral:
Merging this now, to be continued with JATS snippets (#774).

Issue #763, opened 2018-09-11
Adding a new author/editor breaks


michael:
The bio property must be set (a bio node needs to be created).

Integral:
pls check #775

Integral:
This is merged, but hopefully will changed with a JATS snippets.

PR #775, opened 2018-09-11
Insert person fixes

Integral:
## Why

Person insertion is not working as described in #763 

## What

This PR contains a transaction for creation a new person with required biography property.


Integral:
Merging this now, to be continued with JATS snippets (#774).

Issue #798, opened 2018-09-12
Saving is broken


michael:
Regression of cb refactor....

oliver----:
Fixed via #802

PR #802, opened 2018-09-12
Fix save(cb) regression.

oliver----:
## Why

see #798 

## What

Fixes a regression in TextureWebAppChrome


oliver----:
Checked the results on travis, they are green. However github does not seem to receive the result.

PR #804, opened 2018-09-12
WIP: Multi image upload

Integral:
## Why

We want to be able to upload multiple images.


oliver----:
Just noticed, that now we are creating too many empty paragraphs now. We should only do this between to inserted figures, i.e. do it before insertBlockNode() but not for the first one.

oliver----:
@Integral I let you continue here

PR #805, opened 2018-09-12
Fix electron build

oliver----:
Fixing regressions in Electron integration.


oliver----:
Merging this as it is good for a first iteration.

PR #807, opened 2018-09-13
Fix model is empty

oliver----:
## Why

Still we see e.g. an empty biography in an author card, while this is not a required field. 

## What

This adds `isEmpty()` to `ChildModel` and `NodeModel`.


oliver----:
Since yesterday the codecov integration is not working anymore.

Issue #825, opened 2018-09-13
Select IsolatedNodes(e.g. a Figure) when clicking somewhere inside.


oliver----:
After disabling the 'blocker' we want to be able to select the node easily, similar to how Cards work in the Metadata Editor.

![texture-selecting-a-card](https://user-images.githubusercontent.com/284099/45511289-dd470380-b79c-11e8-80f4-85e70fe33587.gif)

oliver----:
Done via #826

PR #826, opened 2018-09-13
Select isolated node

oliver----:
## Why

See #825 

## What

Enables a hook on IsolatedNodes that selects the node on click.
Plus added a test to make sure that this works well together with TableComponent.


oliver----:
@michael this is so much better 😄

PR #865, opened 2018-09-17
Fix find and replace

oliver----:
## Why

The Replace part of the FindAndReplaceManager was not implemented correctly yet.

## What

Complements the implementation and adds more tests.


oliver----:
texture-find-and-replace

Issue #857, opened 2018-09-16
Empty article information card shouldn't be empty


Integral:
<img width="762" alt="screen shot 2018-09-16 at 12 41 14" src="https://user-images.githubusercontent.com/182010/45595032-e78c1c00-b9ad-11e8-80d9-55e95ade2ea2.png">

Integral:
@michael I know that you suggestion was to use required properties again, but I have a feeling that it's a wrong approach since none of these properties could be required really.

oliver----:
@Integral I agree, let's not abuse the 'Required Properties' feature for that, this would also trigger the validator.
Isn't this just a UI / Representation problem? Just one idea, maybe we should render a message explaining that there have been no article info entered yet?

oliver----:
BTW, I find that it is very inconsistent ATM what cards are considered empty, and what not:

michael:
Fixed via #874

PR #874, opened 2018-09-18
Empty cards

Integral:
We want to display at least 3 fields for each card as described in #857.


michael:
Think the implementation could be improved but merging this for our test build.

Issue #884, opened 2018-09-19
Archive can not be opened again after inserting a figure


oliver----:
Reported by @JGilbert-eLife, need to investigate.

JGilbert-eLife:
More detail - if I add in figures, I cannot reopen my work after saving it and exiting. I get the following error when I try to open my document:

ERROR:ParserError: <parsererror xmlns="http://www.w3.org/1999/xhtml" style="display: block; white-space: pre; border: 2px solid #c77; padding: 0 1em 0 1em; margin: 1em; background-color: #fdd; color: black"><h3>This page contains the following errors:</h3><div style="font-family:monospace;font-size:12px">error on line 2 at column 6: XML declaration allowed only at the start of the document </div><h3>Below is a rendering of the page up to the first error.</h3></parsererror>

oliver----:
Ok thanks!

I need to take a closer look. Seems that we are writing out not even XML.

JGilbert-eLife:
This ended up in the XML:
```

<fig id="_figure-1">
          <label>Figure 1</label>
          <caption id="caption-2646c4a4fe6f964cda52fb5f8f97ae39">
            <title>Figure test</title>
            <p id="p-34c78a015bd3c285969ceb7bd351bb96">Testing </p>
          </caption>
          <graphic id="_graphic-1" mime-subtype="png" mimetype="image" xlink:href="25d841d1be04366121330a126e1b328b.png" />
        </fig>
```
however, the manifest file is now showing two XML declarations!
```

<?xml version="1.0" encoding="UTF-8"?>
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE manifest PUBLIC "RDC-Manifest 1.0" "//Manifest-1.0.dtd">
<dar id="dar-1">
  <documents id="documents-1">
    <document id="manuscript" type="article" path="manuscript.xml" />
  </documents>
  <assets id="assets-1">
    <asset id="25d841d1be04366121330a126e1b328b" path="25d841d1be04366121330a126e1b328b.png" type="image/png" />
  </assets>
</dar>
```

oliver----:
Ok... the manifest is broken.

JGilbert-eLife:
Deleting the second declaration means I can open it fine.

oliver----:
Sweet :)
Thanks!

oliver----:
Know how to fix this.

oliver----:
Working on this in #897

oliver----:
Fixed via #897

PR #897, opened 2018-09-19
Fix issue #884

oliver----:
## Why

See #884 

## What

- Add test revealing the problem
- Fix the issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "lodash": "^4.14.1"
   },
   "peerDependency": {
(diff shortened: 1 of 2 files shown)


oliver----:
Unfortunately this seem to occur only in the browser.

oliver----:
This is happening on the browser:
image

oliver----:
This was caused by a bug in Substance (see https://github.com/substance/substance/pull/1229 for details) which has been fixed in version 1.0.0-preview.99

PR #903, opened 2018-09-19
Find and replace styling

Integral:
## Why

We want to improve style of find and replace widget. See mockup from #862


michael:
Looking good basically. Working on some improvements. E.g. the F+R panel should use flex not absolute positioning, otherwise content slips underneath it. Need to change the HTML structure for doing that. Will push a rebased version soon.

oliver----:
Yeah I thought, flex would be a better approach here.

michael:
Think we are good to merge, regarding styling.

oliver----:
Looks great! thx

Issue #896, opened 2018-09-19
Texture Desktop: Store as .dar files instead of a folder


michael:
As reported by Giuliano:

> Saving does not create a folder / DAR with the article name, it dumps the manifest and main xml in whatever folder you clicked. 
> Texture does not associate itself with XML files in MacOS (another reason why DARs should be archives with their own extension and not just folders)

There's some beauty in using folders. Like that you can always look inside the documents and e.g. use Git to track changes made to some files, not the whole file. However there are of course limitations and confusion related to that approach.

I'd suggest to continue for a while with the folder approach until the Dar spec has fully stabilised. As soon as we introduce a file format (.dar) users will expect to be able to open them in all future versions of Texture.

gmaciocci:
That makes file association impossible on desktop. If you're trying to encourage authors to move away from Word, I'm afraid dumping a bunch of xml files in the target folder (especially without a unique subfolder) won't be doing them any favours. Additionally, saving two documents in the same target folder will overwrite the first one, as you seen not to be giving saved files unique names (I always get manuscript and manifest)

oliver----:
I agree that a folder as target is dangerous and a rather advanced feature. It makes sense only if the user wants to save into a git managed repo, for instance.

Integral:
Couldn't we have both options? E.g. save as DAR by default and optionally
export files into folder?

oliver----:
Probably it is better to move to a zip-file with `.dar` as extension. Format-/version-wise I don't see any difference here, because a folder is not much different to a zip file from our angle.

gmaciocci:
@Integral @oliver---- If I may make a suggestion, I'd explore making saving of new files default to .dar zips, and have an "export XML" as a separate option. a single self contained DAR (and desktop file association) makes it easy to open a file multiple times while working on it, and to share it, whereas XML is more of a final product of the work. 

Please ask for a unique name either way (and create a unique subfolder with that name for the manuscript and the manifest for xml exports).

That said, if you open XML to begin with, your default save is XML, unless you do "save as" which will create a new DAR file with a new name.

gmaciocci:
> you can choose that by creating a folder from the save dialog. 

Not to be difficult, but can you name me another application that saves every file with the same name, two files in fact (one which doesn't even make sense to the author), and dumps them in whatever target folder you pick without prompting you for a name?

[...]

PR #915, opened 2018-09-21
Dar Files

oliver----:
## Why

In the Desktop app we want to load and write `.dar` files instead of folders.
See #896 for discussion.

## What

This PR introduces a Storage implementation, that maps zipped DAR archives into an internal unpacked folder when opening a  `.dar` file, and vice versa when saving the DAR.


oliver----:
@michael I had to go back to Electron 2, there were weird app bundling problems.
I'd like to get the 1.0 release done today.
After that I'd like to do a 1.1 with some polishing, where we could also try the electron upgrade.

Issue #919, opened 2018-09-25
Focus highlight is missing for abstract, figure caption etc.


michael:
Regression, we should fix before releasing 1.0.

Integral:
@michael #920

michael:
Fixed via #920

PR #920, opened 2018-09-25
Repair focus highlighting in manuscript

Integral:
This is a fix for #919.

The reason why it was broken:
The ability to list more than one selector for `:not()` pseudo-class  is experimental and not yet widely supported.


michael:
Thank you @Integral !

PR #930, opened 2018-09-27
Schema cut down

oliver----:
## Why

ATM TextureArticle schema is not strict enough. It allows content which is not yet explicitly supported.
In the past we have allowed that, but we don't want to continue doing that, because we are less aware of
what Texture actually is able to deal with.

## What

- Radical cut down to what is actually supported
- Generating a schema documentation 

I have introduced two RNG extensions that allow to remove an element and all its occurrences from the schema, and to mark elements as 'not-implemented'
`NotImplemented` elements can not be edited in Texture, and byy default these are considered errors during import. To allow evaluations in presence of unsupported elements, it is possible to specify a HTML query parameter like so: `localhost:4000/?allowNotImplemented=true`


oliver----:
Here is the generated documentation: https://github.com/substance/texture/blob/2ad9bbdbdfacfc2486e63fe794baaf42ff9ff722/docs/TextureArticle.md

Issue #948, opened 2018-10-15
Some other footnote bugs


Integral:
- when you click add a new reference in metadata panel other footnote got selected (a new one should be selected)
- empty footnote should take a full width, currently it looks like this:
![screen shot 2018-10-15 at 10 18 33](https://user-images.githubusercontent.com/182010/46935703-a3e80900-d064-11e8-906d-3ec1eda09cfe.png)

Integral:
Fixed in master via #994.

Issue #947, opened 2018-10-11
Can't enter text to newly created footnote


pfeiffer0kristina:
Hi! We are testing texture at the moment for DAI and I have a problem with the new released version when it comes to footnotes. Neither can I fill the field with information, nor can I crossref. I can open up a footnote field in metadata but then nothing more. In the code texture complains about a blank entry in ft resulting in a reload failure of the file....Sorry for asking, maybe I just do not get it but I cant solve by myself. In the previous versionj of texture it was fine! Thank you!

Integral:
Hello!
Current version of Texture designed more for publishers, so idea was that you are editing and adding new footnotes through a metadata mode.
However, as I can see now you can't enter text to a newly created footnote, so this is actually a bug. Thanks for reporting!

pfeiffer0kristina:
Thank you!

Integral:
Fixed in master via #994. Not yet released.

PR #994, opened 2018-10-23
Footnote issues

Integral:
## Why

Currently there are couple of problems with new or empty footnotes, see #947, #948.

## What

This PR contains tests which reveals problems and fixes. I'm manually appending empty paragraph during creation and during jats2internal conversion for empty paragraphs (could imagine that someone would like to open article with empty fn element, why not?). Don't like this solution very much, the alternative could be probably a custom converter, but don't want to do it know. I think we need to collect and analyse more cases and then decide how we want to nail this empty paragraphs problems for containers generally.
Also here is a minor style fix.


Integral:
Thanks for reviewing

PR #1013, opened 2018-11-05
Issue title for article meta

Integral:
## Why

Érudit requirement: Texture should support the editing of `issue title` in the metadata editing.

## What

This PR contains basic support of `issue title` inside `article-meta`.
Now we need to decide whatever do we want to support languages for this element.
And then we will need to decide if we want to support this element inside `element-citation`.

### DIFF
--- a/src/article/InternalArticleSchema.js
+++ b/src/article/InternalArticleSchema.js
@@ -26,6 +26,7 @@ ArticleRecord.schema = {
   type: 'article-record',
   volume: STRING,
   issue: STRING,
+  'issue-title': STRING,
   fpage: STRING,
   lpage: STRING,
   pageRange: STRING,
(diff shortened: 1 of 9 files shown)


src/article/InternalArticleSchema.js
```diff
@@ -26,6 +26,7 @@ ArticleRecord.schema = {
   type: 'article-record',
   volume: STRING,
   issue: STRING,
+  'issue-title': STRING,
```
oliver----: @michael are you confident that the title is plain text?

oliver----:
Beyond my question this PR is fine.

oliver----:
The JATS spec allows rich-text in the `<issue-title>`.

Integral:
Ok, I will do that

fabiobatalha:
@oliver---- @Integral 

Sorry I have just realized that even we want to have rich-text in issue-title, JATS will not allow us to do that.

https://jats.nlm.nih.gov/publishing/tag-library/1.2d2/element/issue-title.html

JATS do not allow rich-text in issue-title

oliver----:
This is good for a first iteration.

Issue #966, opened 2018-10-17
Ability to add preformat blocks


oliver----:

michael:
Maybe we call this preformat as in JATS.

```xml
<preformat preformat-type="code">
def skip_test_launch_environ():
    host = Host()

    assert len(host._binders) == 0

    ioloop = IOLoop.current()

    async def run():
        result = await host.launch_environ('https://github.com/org1/repo1')
        ioloop.call_later(0.5, ioloop.stop)

    ioloop.add_callback(run)
    ioloop.start()

    assert len(host._binders) == 1

    binder = next(iter(host._binders.values()))
    assert 'request' in binder
    assert binder['request']['url'] == 'https://mybinder.org/build/gh/org1/repo1/master'
    assert binder['events'] == []
</preformat>
```

I guess we should also assume CDATA here. E.g. when you want to provide HTML code.

Integral:
@michael, I would say that we should use ```code```. We already have it, but a bit unfinished. For inline code we are using `monospace`.
Did I miss something?

Integral:
It looks to me that it might make sense to have a references to code blocks... @michael?

oliver----:
@Integral interesting general question.
@michael I guess there are more things that could or should be referencable, e.g. sections etc.

Integral:
Could I close this issue? Or do we need a references now?

oliver----:
This is good for now.

PR #1038, opened 2018-11-18
Preformat

Integral:
## Why

As requested in #966 we want to support preformat blocks.

## What

This PR contains necessarily schema changes for using `preformat` tag as well as converters, styles and command. It also removes any ability to use `code` tag since we are not allowing usage of few tags per use case.


oliver----:
This is looking good.

@Integral Should we do something about the anomality of `parseSnippet` in presence of CDATA, i.e. do you think this is inconsistent?

Integral:
@oliver---- good question
It's not very consistent indeed.
I think that we should do some iteration on CDATA, there is a bug there, which I described here: https://github.com/substance/substance/issues/1233, it's getting important with block formulas.
Generally speaking I think that as an API user I would expect to get CDATA content in the same way as regular textual data, without any additional calls. If it's possible.

oliver----:
I'll try this out and then merge.

oliver----:
> It's not very consistent indeed.
> I think that we should do some iteration on CDATA, there is a bug there, which I described here: [substance/substance#1233](https://github.com/substance/substance/issues/1233), it's getting important with block formulas.
> Generally speaking I think that as an API user I would expect to get CDATA content in the same way as regular textual data, without any additional calls. If it's possible.

I will track this in Substance

Integral:
Thank you!

oliver----:
https://github.com/substance/substance/issues/1234

oliver----:
rebased.

Issue #1018, opened 2018-11-06
Add support for inline graphics


JGilbert-eLife:
eLife does allow inline graphics as distinct from main figures. We are reconsidering their use in general but we have examples in the archive of this being used to place images within tables.

```
<sec id="s2-6-4">
                    <title>Inline graphics</title>
                    <p> Here is an example of pulling in an inline graphic <inline-graphic mimetype="image" mime-subtype="jpeg" xlink:href="elife-00666-inf001.jpeg"/>.</p>
                </sec>
```

Integral:
<img width="636" alt="screen shot 2018-11-18 at 19 03 03" src="https://user-images.githubusercontent.com/182010/48675091-71e72c80-eb65-11e8-9103-62965935d0d0.png">

PR #1039, opened 2018-11-19
Inline graphics

Integral:
## Why

There was a request #1018 for inline-graphics support.

## What

This PR contains all the schema changes, tools and commands for inline graphics implementation.


src/article/shared/InlineGraphicComponent.js
```diff
@@ -0,0 +1,19 @@
+import { NodeComponent } from '../../kit'
```
oliver----: I think that the implementation of GraphicComponent.js and InlineGraphicComponent.js should be assimilated.

Integral: Agree, but first we need to think what to do with displaying errors. 
E.g. when there is a file missing for a figure we are displaying a big red warning, but in case of a small inline graphic... I do not know.

oliver----:
@Integral please add a test for inline-graphics here: https://github.com/substance/texture/blob/c8d62bd883ab6df4be95df3b65fe8d251a2952ce/test/ManuscriptEditor.test.js#L7-L21

src/article/shared/InlineGraphicComponent.js
oliver----: Right. @michael pls

Integral:
Ok, cool. Let's put this on hold for some time.
2 general things left:
1) Test (i think we also need a test for missing file)
2) Missing file warning implementation (would be great to have the same implementation for inline and non-inline cases)

Thanks

Integral:
Hey @michael!
I think that for missing inline graphics we should use some kind of icon, like this one: 
Maybe the combination of icon, red color and title could do the trick, what do you think?
Another question which is bothering me, should we restrict the height of inline graphic? I think we need a limit here to avoid big space between two lines, like here:
screen shot 2018-11-24 at 17 16 33

Integral:
@oliver---- I think it is finished now. This is what I did for errors within inline graphic:
screen shot 2018-11-27 at 10 05 23
There is no test for converter since there is no dedicated converter for inline graphic, so it is covered by general JATS import/export and new inline graphic insertion test.

oliver----:
> Note that I put some tools inside `insert` dropdown menu since there is no space left in toolbar. That also required adjustment of some tests. E.g. this PR most be merged before others tool-related PRs like the one with `block formula`. The latest should be rebased and fixed according to toolbar changes.

The toolbar layout needs to be revisited. In fact all insert tools should go there, because the differences are too subtle to be able to find icons for it (e.g. figure vs inline graphic)

oliver----:
Looking good. Will merge.

PR #1041, opened 2018-11-21
Affiliation name -> Institution.

Integral:
## Why

As described in #924 we want to change label of affiliation.name from 'Name' to 'Institution'.

## What

Since `organisation` entity have property `name` we can't just change it's label. Here I am changing property to `institution` as well as converters and renderers and then change label.


oliver----:
rebased.

oliver----:
Looking good.

PR #1042, opened 2018-11-21
WIP: block formulas

Integral:
General work on `<disp-formula>` and block formula implementation.


src/article/editor/EditorPackage.js
```diff
@@ -213,10 +234,12 @@ export default {
     config.addLabel('insert-table', 'Table')
     config.addIcon('insert-table', { 'fontawesome': 'fa-table' })
 
+    config.addTool('edit-block-formula', EditDispFormulaTool)
```
oliver----: We should use a consistent name, either `disp-formula` or `block-formula`
@michael pls your preference

oliver----:
Beyond my comment, this is looking good. Of course we wait for the substance bug being fixed.

src/article/editor/EditorPackage.js
oliver----: My personal preference is actually 'block-formula' similar to 'block-quote'

oliver----: In the AMS extension for Lens we used 'equation' here.

Integral:
Still, this is not ready yet.
There are few problems here:
1) The coordinates of overlay editor is somehow computed wrong, it's always rendered much lower than is should be. Also using the same small input as we have for inline formula is not a good idea I think.
2) There is a strange bug, somehow after curtain sequence of clicks selection and viewport are moving to the beginning of manuscript...

oliver----:
Ok, good to know.

oliver----:
@Integral the validation issue is resolved.

Integral:
Great! Thank you very much!

oliver----:
Last updates looking good.

Integral:
Waaait, there will be tests for converters :) And then I will need to fix a bug with overlay input (for some reasons block formula inputs are much lower than formula itself)

Integral:
@oliver---- re overlay input problem, there are problems with selectionRect height which is growing fast for complex katex markup.
For instance in this case:
screen shot 2018-11-28 at 22 19 13
selectionRect height ~ 500px
Do you have any ideas what we can do with it?

oliver----:
@Integral I need to take a look. With MathJax we had a similar case where we needed to use a specific child-element for the bounding rect.

oliver----:
Is this the last issue in this PR, i.e. that the popup is not positioned correctly?

oliver----:
I guess we need to change the `_getSelectionRect()` implementation in `AbstractScrollPane`.
In case of a NodeSelection (as in this particular case) the bounding rect be the one of the isolated node.

oliver----:
@Integral I changed the logic for selectionRect computation, to use the node's element instead of the Window selection in case of NodeSelection. Please try out. You can merge when happy.

Integral:
Thanks, I will take a look

PR #1055, opened 2018-11-29
Preformat tests

Integral:
Since I add some helpers... @oliver---- just in case, take a look please!


oliver----:
Looking good 👍

Issue #556, opened 2018-06-05
Table footnotes


JGilbert-eLife:
In addition to a title and a caption, eLife supports footnotes in tables. These can be linked to the content of the table but do not have to be.

```
<table-wrap>
. . .
<table>
. . .
<th rowspan="2">Molar ratio of lipid:protein in RPL reactions<xref ref-type="table-fn" rid="tablefn1">*</xref></th>
. . .
                       </table>
                        <table-wrap-foot>
                            <fn id="tablefn1"><label>*</label>
                                <p>Footnotes can be used to highlight properties of data reported in a table such as statistical significance. They are separate from
                                    the table caption and appear afterwards. They are hyperlinked to allow easy navigation. Footnotes in tables use the same standard
                                    set of symbols used for authors footnotes.</p>
                            </fn>
                            <fn id="tablefn2"><label>†</label>
                                <p>Authors are fully allowed to cite references and figures in tables. There is no difference in citation style between the main text and tables.</p>
                            </fn>
                            <fn><p>Order: Designated footnotes (e.g. &#x2a;, &#x2020;, &#x2021;, &#167;, &#035;, &#182;, &#x2a;&#x2a;, and so on), p value footnotes (&#x2a;p, &#x2a;&#x2a;p, &#x2a;&#x2a;&#x2a;p),
undesignated footnotes and abbreviations.</p>
                            </fn>
                        </table-wrap-foot>
</table-wrap>

```

michael:
I'm not convinced we need specific table footnotes. You could still use footnotes in captions but they would end up in the global fn-group. At eLife you could decide to render them in-place, but I think this should not be a main functionality of Texture. It complicates the XML structure and UX.

Melissa37:
Humm, not sure about that, footnotes are quite specific and different from the caption in my opinion. Would be interested to hear what others think - I appreciate the idea, but would appreciate other thoughts from other publishers

JGilbert-eLife:
Just writing this out for clarity (and to check my own thinking) --

The specific use for footnotes is cross-linking into the main body of the table. So, for example, if you have a footnote

` <fn id="tablefn1"><label>*</label><p>This is a definition of what * indicates in the table.</p></fn>`

you can then call this out in the body of the table with an xref:

`<xref ref-type="table-fn" rid="tablefn1">*</xref>`

This is of course a key part of the way in which tables are presented in journal articles, allowing authors to append extra information. For example, they might want to draw a distinction between some values in a row:

[...]

PR #1057, opened 2018-12-03
Table footnotes

oliver----:
## Why

Implementation of Feature 'Table Footnotes'. 

Related:
- #556

## What

- [x] Extend kitchen-sink **(O)**
- [x] Extend TextureArticleSchema to allow for `<table-wrap-foot>`
- [x] Extend `TableFigure` to allow for 0-n footnotes (-> children)
- [x] Implement Importer and create footnotes and populate `tableFig.footnotes`
- [x] Implement a label generator for table footnotes **(O)**
- [x] Refactor internal xref collection API for XrefEditDialog **(O)**  
- [x] Display footnotes and footnote references in TableFigure
- [x] Implement dialog for table-footnotes so that it allows only to choose from table footnotes
- [x] InsertFootnoteCrossReferenceCommand for both regular and table footnotes according to detected scope
- [x] Determine xpath for table selections (CustomSelection) 
- [x] Implement an InsertFootnoteCommand that creates a footnote in the according scope
- [x] Add InsertFootnoteTool to both views (currently only metadata)
- [x] After creating the first footnote xref, table- footnotes do not update anymore **(O)**
- [x] Fix TableFootnoteManager issues (e.g. no label for xrefs in caption) **(O)**
- [x] Add tests for table-footnote label generator **(O)**
- [x] Order footnotes in order of cross-refs (similar to how references are ordered) **(O)**
- [x] Implement exporter that stores created footnotes into `<table-wrap-foot>` **(O)**
- [x] Add a command and tool to remove a footnote (enabled when cursor inside fn)
- [x] Add tests for table-footnote interaction (create, delete, x-reference)
- [x] Fix table editor in Metadata mode
- [x] Add labels for table figures ('Label:', 'Title:', 'Caption'; 'Footnotes:') and regular figures styled in a same way as we have it in manuscript


oliver----:
Note: please introduce a TableFigureModel (extends FigureModel) and a TableFigureComponent (extends FigureComponent)

oliver----:
@Integral I have started a first iteration on the label generator (WIP)

oliver----:
If you need the label for rendering, there is `tableFigure._tableFootnoteManager` which has `getLabel(pos)`

oliver----:
For the EditXrefDialog I will have to do some refactorings. The current approach is not sufficient.

Integral:
@oliver---- do we want to use collections for managing table footnotes? E.g. one collection for each figure? I see that with footnote manager we are already going this way, and I think that it we should do that the same with collections...

Integral:
Nice. Also it would be easier for us to manage it.

oliver----:
UI needs improvement: #1069

oliver----:
Quick-fixed the Xref target resolver. At some point this needs a refactor.

Integral:
screen shot 2018-12-05 at 13 37 09
We should do an iteration on metadata table. There are no fn renderers and wrong css classes now. Also tools should be available there. My suggestion would be to do it separately closer to test phase.

oliver----:
@michael we need your action regarding a potential requirement for re-ordering of (table) footnotes.

src/article/editor/RemoveFootnoteCommand.js
```diff
@@ -0,0 +1,15 @@
+import RemoveItemCommand from './RemoveItemCommand'
+
+export default class RemoveCollectionItemCommand extends RemoveItemCommand {
```
oliver----: The file is called RemoveFootnoteCommand. Would like to rename the file or rename the class?

oliver----: I'd say, go for RemoveFootnoteCommand.
When we introduce a similar one we gonna extract a shared impl.

oliver----:
@Integral very nice! Thanks for the tests!

oliver----:
Perfect! 👏

oliver----:
Let's get this merged! 🍾 🎉 🎆

Integral:
Wonderful!

Issue #935, opened 2018-09-27
We should be able to insert Reference Citations inside table cells, caption and title


lpanebr:
Texture currently does not allow this, and yet it's something we see every day.

lpanebr:
ow, and inside table footnotes also (when we have them #556).

Melissa37:
Agree with this completely - and figure legends.

JGilbert-eLife:
Ah, sorry! Accidentally part-duplicated this with the wider case of https://github.com/substance/texture/issues/970 (formatting, hyperlinks and so on also don't work in table cells).

The issue in this ticket also applies to figure titles.

oliver----:
Almost done. Only for the title it is still missing.

oliver----:
Solved via #1077

PR #1077, opened 2018-12-07
Enable xrefs in Figure and Table  title

oliver----:
## Why

See #935 

## What

- Adapts the internal model to allow for xrefs
- Adds tests for creating reference citations


oliver----:
In another iteration we should solve #1077 .

Issue #1012, opened 2018-11-05
Support for figure supplements


michael:
We want to support "Stacks of Figures". At eLife there's one main figure and any number of figure supplements. The desired tagging for this in JATS is as follows:

```xml
<fig-group content-type="stack">
<fig id="fig3">...</fig><!--main figure-->
<fig id="fig3s1" specific-use="child-fig">...</fig><!--supplement 1 -->
<fig id="fig3s2" specific-use="child-fig">...</fig>
<fig id="fig3s3" specific-use="child-fig">...</fig>
</fig-group>
```

We added `content-type="stack"` to be able to distinguish this from another type of grouping ?`content-type="panels"`, which represents a single figure, consisting of multiple individual panels. (to be described in a separate issue).

We will have to adjust automatic labelling and detect such groups. Instead of Figure 3, Figure 4, Figure 5 we would need to generate "Figure 3, Figure 3 - supplement 1, Figure 3 supplement 2, ...". Is this general enough for a Stack feature?

michael:
@dalapeyre there is just one use-case, we may also have which is a "stack of multi-panel figures". Therefore we would need nesting of fig-group like:

```xml
<fig-group content-type="stack">
  <fig-group content-type="panels" id="fig1">...</fig-group>
  <fig-group content-type="panels" id="fig2s1">...</fig-group>
</fig-group>
```

This is not a pressing issue, but maybe it would make sense to allow nesting fig-group elements in JATS. Up for discussion...

michael:
eLife Example XML (of current corpus, not Dar):

```xml
<fig-group>
                        <fig id="fig2" position="float">
                            <object-id pub-id-type="doi">10.7554/eLife.00666.008</object-id>
                        <label>Figure 2.</label>
                        <caption>
                            <title>Figure with figure supplements. In the PDF this asset box will take full column width.</title> 
                            <p>This is the basic information provided about an article. <xref ref-type="fig" rid="fig1">Figure 1
                            </xref> shows an expanded view (<xref ref-type="bibr" rid="bib26">Kok et al., 2015</xref>; <xref ref-type="bibr" rid="bib28">National Institute of Mental 
                                Health, 1990</xref>).</p>
                        </caption>
                            <graphic mimetype="image" mime-subtype="tiff" xlink:href="elife-00666-fig2.tif"/>
                    </fig>
                        <fig id="fig2s1" position="float" specific-use="child-fig">
[...]

PR #1084, opened 2018-12-10
Figure Groups

oliver----:
## Why

Implementation of feature 'Figure Groups' (multi-panel figures, figure supplements)

Related:
- https://github.com/substance/texture/issues/1012

## What

- [x] Extend the kitchen-sink **(O)**
- [x] Introduce FigurePanel in internal schema **(O)**
- [x] Change Figure to `{ panels: CHILDREN(figure-panel) }` **(O)**
- [x] Refactor importer for Figure and TableFigure **(O)**
- [x] Import of FigurePanels **(O)**
- [x] Introduce FigurePanelComponent **(O)**
- [x] Refactor TableFigureComponent **(O)**
- [x] Implement FigureComponent to display only one panel at a time (manuscript view) **(O)**
- [x] Adapt the FigureManager implementation to generate appropriate labels, e.g. Figure 1, Figure 1A **(O)**
- [x] Add AddPanelTool to contextual toolgroup
- [x] Implement FigureComponent to display all panels unfolded (metadata view)
  - Challenge is that FigurePanels should be represented as individual cards, so that we do not destroy our flat card UI
- [x] Introduce AddPanelCommand (scoped to figure)
- [x] Introduce command and tool to remove the current panel
- [x] Implement FigureComponent for the Manuscript (following the mockup in the requirements document)
- [x] Introduce command and tool to change the position of the current panel (move up, move down)
- [x] Introduce command and tool to replace the image of the current panel
- [x] Add tests (see below)

## Mockups

- User Interface for a Figure with multiple sub-figures in Manuscript View: 

image

## Tests
- [x] Add a panel to a figure
- [x] Remove a panel from a figure
- [x] Remove a figure with multiple panels
- [x] Undo the removal of a figure with multiple panels
- [x] Change the order of panels
  - [x] Move up
  - [x] Move down
- [x] Replace image of a panel
- [x] Reference a sub-figure
- [x] Reference a sub-figure and change order of panels
- [x] Reference a sub-figure and then remove panel
- [x] Meta-data view:
  - [x] check that each panel is in individual card
  - [x] click on a figure panel and remove it
  - [x] and move up figure 2b


src/article/editor/EditorPackage.js
```diff
@@ -134,6 +136,9 @@ export default {
       nodeType: 'fig',
       commandGroup: 'additional'
     })
+    config.addCommand('add-panel-figure', AddFigurePanelCommand, {
```
oliver----: Better call this 'add-figure-panel' (like in the Command name)

src/article/editor/EditorPackage.js
```diff
@@ -253,6 +258,10 @@ export default {
     config.addLabel('insert-fig', 'Figure')
     config.addIcon('insert-fig', { 'fontawesome': 'fa-image' })
 
+    config.addTool('add-panel-figure', AddFigurePanelTool)
+    config.addLabel('add-panel-figure', 'Add Panel Figure')
```
oliver----: We want to call the label 'Add Sub-Figure'

src/article/shared/AddFigurePanelCommand.js
```diff
@@ -0,0 +1,39 @@
...
+  getCommandState (params, context) {
+    return {
+      disabled: this.isDisabled(params)
+    }
+  }
+
+  execute (params, context) {
+    const collection = this._getCollection(params, context)
+    const files = params.files
+    collection.addItem({})
```
oliver----: I think you should not add twice. I.e. add only one new item to the current panel collection.

src/article/shared/AddFigurePanelCommand.js
```diff
@@ -0,0 +1,39 @@
...
+    let api = context.api
+    if (files.length > 0) {
+      api._insertFigurePanel(files)
```
oliver----: ... see previous comment: I think we should use the collection's addItem(), but only do that once.

src/article/shared/FigurePanelCommands.js
```diff
@@ -25,37 +25,21 @@ class BasicFigurePanelCommand extends Command {
...
 
 export class RemoveFigurePanelCommand extends BasicFigurePanelCommand {
   execute (params, context) {
     const figureModel = this._getFigureModel(params, context)
-    const panels = figureModel.getPanels()
-    const index = this._getCurrentPanelIndex(figureModel)
-    const panel = panels.getItemAt(index)
-    let api = context.api
-    api._removeFigurePanel(panel, panels)
+    figureModel.removePanel()
```
oliver----: Yes. That's what I thought too in general. Use the models as entry points, and in the models use the API for the nasty stuff.

Issue #1016, opened 2018-11-06
Add support for supplementary-material in caption


JGilbert-eLife:
eLife allows two types of `<supplementary-material>` within `<caption>` for figures, tables and videos - source data and source code.

(This is a separate but related use of supplementary-material at the article level, see #1015 )

Examples:

Source data for a table:

```
                    <table-wrap id="table3" position="float">
                        <object-id pub-id-type="doi">10.7554/eLife.00666.005</object-id>
                        <label>Table 3.</label>
                        <caption><p></p>
                            <p><supplementary-material id="table3sdata1">
                                <object-id pub-id-type="doi">10.7554/eLife.00666.006</object-id>
                                <label>Table 3—source data 1.</label>
                                <caption><title>Representative curves of steady-state kinetic analyses for each IGF1R protein characterized.</title>
                                    <p>Each data point was performed in duplicate and is shown separately.</p>
                                </caption>
                                <media mimetype="application" mime-subtype="xlsx" xlink:href="elife-00666-table3-data1.xlsx"/>
                            </supplementary-material>
                            </p>
                        </caption>
                    . . . . . . .

```

Source data for a figure supplement (see: #1012) 

```
<fig id="fig3s1" position="float" specific-use="child-fig">
                            <object-id pub-id-type="doi">10.7554/eLife.00666.012</object-id>
                            <label>Figure 3—figure supplement 1.</label>
                            <caption>
                                <title>Title of the figure supplement</title>
                                    <p>
                                        <supplementary-material id="fig3sdata1">
                                            <object-id pub-id-type="doi">10.7554/eLife.00666.013</object-id>
                                            <label>Figure 3—figure supplement 1—source data 1.</label>
                                            <caption>
                                            <title>Title of the figure supplement source data.</title>
                                            <p>Legend of the figure supplement source data.</p>
                                            </caption>
                                            <media mimetype="application" mime-subtype="xlsx" xlink:href="elife-00666-fig3-figsupp1-data1.xlsx"/>
                                        </supplementary-material>
                                    </p>
                            </caption>
                            <graphic mimetype="image" mime-subtype="tiff" xlink:href="elife-00666-fig3-figsupp1.tif"/>
                        </fig>
```

Source code for a video:

```
<media mimetype="video" mime-subtype="mp4" id="video1" xlink:href="elife-00666-video1.mp4">
                        <object-id pub-id-type="doi">10.7554/eLife.00666.016</object-id>
                        <label>Video 1.</label>
                        <caption>
[...]

PR #1087, opened 2018-12-18
Feature: Supplementary files

oliver----:
## Why

See #1016, for instance.

## What

Supplementary files are represented as block-level elements, i.e. they can be placed in the manuscript body, or in an appendix, or in a figure caption. Intentionally we decided to allow this -- as it is defined in JATS -- as a general block-level element because this gives the author the most freedom in how to use these. E.g. they can be used anywhere in the manuscript body, or in a figure caption. Supplementary files can be referenced like other resources such as figures.

- [x] Extend the kitchen-sink (O)
- [x] Introduce SupplementaryFile model (O)
- [x] Add importer and exporter (O)
- [x] Implement label generator (simple) (D)
- [x] Implement UI Component (D)
- [x] Implement a command and tool for uploading and inserting supplementary file (D)
- [x] Ability to reference a supplementary file (D)
- [x] Ability to update an existing supplementary file (D)

Tests:
- [x] create a supplementary file (D)
- [x] remove a supplementary file (D)
- [x] reference a supplementary file (D)
  - [x] then insert another supplement before (reference should have updated)
  - [x] remove the referenced supplement (reference should be broken)
- [x] update an existing supplementary file
  - until we have better means to check if the file has actually changes, we should just check that the code does not throw

Not in this PR:
- Scoped labels, e.g. 'Figure 1 - Supplementary File 1.'

Mockups:

image


src/article/converter/r2t/SupplementaryFileConverter.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+    if (labelEl) {
+      node.label = labelEl.text()
+    }
+    node.href = el.getAttribute('xlink:href')
+    node.legend = importer.convertElement(captionEl).id
+  }
+
+  _getContent (el) {
+    return findChild(el, 'tex-math')
```
Integral: deadcode

oliver----: Thx!

Integral:
I think it's finished.

test/SupplementaryFile.test.js
```diff
@@ -117,3 +117,26 @@ test('Supplementary File: reference a file', t => {
...
+  loadBodyFixture(editor, SUPPLEMENT_FILE)
+
+  const replaceSupplementaryFileToolSelector = '.sc-upload-supplementary-file-tool.sm-upload-tool'
+  editorSession.setSelection({
+    type: 'node',
+    nodeId: 'sm1',
+    surfaceId: 'body',
+    containerId: 'body'
+  })
+  const replaceSubFigureImageTool = editor.find(replaceSupplementaryFileToolSelector)
```
oliver----: should be called 'replaceSupplementaryFileTool'

src/article/shared/UploadTool.js
```diff
@@ -3,14 +3,17 @@ import { ToggleTool } from '../../kit'
...
-    const fileType = this.getFileType()
     const isMultiple = this.uploadMultiple()
     const input = $$('input').attr({
-      'type': 'file',
-      'accept': fileType
+      'type': 'file'
     }).ref('input')
       .on('change', this.onFileSelect)
 
+    if (!this.acceptAllFileTypes) {
```
oliver----: This looks like a regression. Either define it as a property, or call it as a function

oliver----: Spotted it, because accelAllFileTypes was never called.

PR #1089, opened 2018-12-28
Changing the Menu Layout (Iteration I)

oliver----:
## Why

The current menu layout is inconsistent and thus confusing, and does not scale well.

## What

In this iteration we introduce a more canonical menu layout, using the same configuration
in manuscript and metadata view

```
## Document Tools (tool group, flat icons)
Undo
Redo
Save

## Primary Annotations (tool group, flat icons)
Bold
Italic
Link

## Insert (drop down menu)
Figure
Table
Image
Blockquote
Equation
Footnote
Reference
Supplementary File
--- Inline ---
Math
Inline Graphic
Link
Citation
Figure Reference
Table Reference
File Reference
Footnote Reference
---- Metadata ----
Author
Editor
Group
Affiliation
Award
Keyword
Subject

## Format (drop down menu)
Bold
Italic
Monospace
Subscript
Superscript
Underline
Overline
Strike-Through
Small Caps

## Switch text types (dropdown, hide disabled)
## Table tools (dropdown, hide disabled)
## List tools (dropdown, hide disabled)
## ... etc. (for every 'context' a dropdown that hides when disabled)
```

> Notice, that we use dropdowns for the contextual tools in this iteration. It is already there, and fulfills the principal requirements.

- [x] Bring back custom tools e.g. for InsertInlineGraphic
- [x] Add switch text-type commands for list types, and activate the list-tools dropdown only when in list. **(O)**
- [x] In EditorPackage and MetadataPackage, replace `config.addAnnotationTool()` with `config.addCommand(...)` plus necessary label and icon separately
- [x] Add missing labels to ArticleToolbarPackage
- [x] Add missing icons to ArticleToolbarPackage
- [x] Add missing contextual dropdowns (figure, supplementary, collection(?))
- [x] Fix regressions in Tests
  - [x] AddEntity.test
  - [x] AddReference.test
  - [x] Annotations.test
  - [x] CrossReference.test
  - [x] EditReference.test
  - [x] Figure.test
  - [x] Footnotes.test
  - [x] ManuscriptEditor.test
  - [x] Table.test **(O)**
  - [x] SupplementaryFile.test


oliver----:
I have done a first iteration: see https://github.com/substance/texture/blob/26a247b9f547c69716e35d8fce117fb0c58e6f34/src/article/shared/ArticleToolbarPackage.js

Some explanations:

I want to decouple Toolbar/Tools, and Commands a little, and also generalize the labels and icons.
This means, a command is registered without the notion of an actual tool: such as in https://github.com/substance/texture/blob/26a247b9f547c69716e35d8fce117fb0c58e6f34/src/article/editor/EditorPackage.js#L340-L344. The command should have a self-explaining name, instead of 'bold' it should be called 'annotate-bold' instead (or maybe 'toggle-bold'). In the Toolbar we link a tool to the command like this: https://github.com/substance/texture/blob/26a247b9f547c69716e35d8fce117fb0c58e6f34/src/article/shared/ArticleToolbarPackage.js#L20

Using an explicit label and icon here to generalize labels and icons, and bring us away from the undesired effect that a command name requires a label. Instead the Toolbar designer chooses how to represent the tool.

As next steps we should:
- apply the tool/command decoupling to the other annotation tools, as well to other existing tools
- add labels and icons in a generalized way as in the above example.

oliver----:
The challenge of contextual toolgroups is solved using dropdowns for now. This is not optimal but possible without introducing new mechanisms.

PR #1093, opened 2019-01-09
Figure Metadata Fields (Iteration I)

oliver----:
## Why

Custom fields and values can be used to capture structured scientific and experimental metadata for dataset indexing.

## What

In this first iteration we add a metadata table to figures (sub-figures) where each row represents a metadata field, and first column the field name, and the second column contains a comma separated list of values.

Tasks:
- [x] provide JATS sample and adapt kitchen-sink **(O)**
- [x] extend the model **(O)**
- [x] implement importer and exporter **(O)**
- [x] extend the UI component (display field table) (D)
- [x] add ability to add, remove, and reorder fields (D)

Tests:
- [x] create a new field
- [x] remove a field
- [x] reorder fields

Mockup:

image


src/article/editor/EditorPackage.js
```diff
@@ -265,6 +266,9 @@ export default {
       nodeType: 'superscript',
       commandGroup: 'formatting'
     })
+    config.addCommand('add-custom-metadata-field', AddCustomMetadataFieldCommand, {
```
oliver----: I'd like to define commands in alphabetic order. Less git conflicts, and more clear how to do it.

Integral: 👍

src/article/ArticleAPI.js
```diff
@@ -437,4 +437,37 @@ export default class ArticleAPI {
       documentHelpers.insertAt(tx, [figure.id, 'panels'], pos, node.id)
     })
   }
+
+  _getSurfaceId (node, path, viewName) {
```
oliver----: You could add a comment explaining your feelings about this implementation, and what the basic idea of the approach is

oliver----: e.g.: 
```
HACK: determining proper surfaceId in a hard-coded way.
Using templates for different use cases to derive ....
```

oliver----: The more you think a method is, the more important is a comment.

Issue #1085, opened 2018-12-11
Change terminology from figure 'caption' to 'legend'


oliver----:
I have noticed that we are using the term figure 'caption' probably not correctly. It would be more common to name the fields in a Figure 'label', 'title', and 'legend'.

JGilbert-eLife:
The `<caption>` element in JATS captures the combined title (`<title>`) and legend (One or more `<p>` following `<title>`). It does not contain `<label>` which is applied outside `<caption>` as a direct child of the object being labelled e.g. `<fig>`.

Therefore it's more accurate to say that a figure would have a label and a caption, with the caption being the composite of the title and legend.

I think that probably holds true for most colloquial uses of 'caption' since it's the structure reflected in JATS. It refers specifically to the descriptive text accompany an object whereas 'label' is the human-readable identifier for the item. 

Not sure how much that really matters for the Texture display though!

oliver----:
Just don't want to use confuse with using terms incorrectly.
After what you said, it would still be better to name fields 'label', 'title', and 'legend'.
Correct?

JGilbert-eLife:
Absolutely, and apologies for not being so succinct!

oliver----:
Thanks! 😄

oliver----:
Done.

PR #1100, opened 2019-01-18
Legend instead of captions

Integral:
## Why

As discussed in #1085 we wanted to use rather `legend` terminology for what was called `figure `caption` before.

## What

This PR contains all necessarily changes to perform renaming and also I've change a name of supplementary file `caption` as well, since we already using `legend` term even on model level.


oliver----:
I will fix regressions from rebase, and the merge.

PR #1105, opened 2019-01-21
Remote supplementary files

oliver----:
## Why

In addition to the existing ability to upload files as ‘Supplementary Files’ it is necessary to be able to link an existing, hosted file as a Supplementary File.
In this iteration we change the UX for adding a Supplementary File, to show a prompt first with two options, one to upload a file, and one to enter a URL

## What

Tasks:
- [x] Add a supplementary file with remote URL to kitchen-sink
- [x] Adapt SupplementaryFile model to have a new property SupplementaryFile += { remote: BOOLEAN }
- [x] Change importer detecting remote link in attribute ‘xllink:href’ and setting remote=true|false accordingly
- [x] Implement a AddSupplementaryFileWorkflow following the Mockup in https://projects.invisionapp.com/d/main/default/#/console/16490835/341955433/preview
  - Note: this should be similar to AddReferenceWorkflow
- [x] Update SupplementaryFileComponent to show an editable href with different label (see mockup)
- [x] Add contextual tool that allows to download the associated file (local and remote)
- [x] Fix tests
- [x] Fix file download for local assets **(O)**
- [x] Add test for downloading local supplementary file

Mockup:

image


src/article/shared/InputActionComponent.js
```diff
@@ -0,0 +1,59 @@
+import { Component } from 'substance'
+
+export default class InputActionComponent extends Component {
```
oliver----: Please add a description here what this Component is made for, as it is not obvious.

oliver----: Also, the name would indicate that this is an 'InputAction', which IMO is not what it is. Let's find a better name.

oliver----:
@Integral please rebase onto latest master

src/article/editor/InsertNodeFromWorkflowCommand.js
```diff
@@ -0,0 +1,12 @@
+import InsertNodeCommand from './InsertNodeCommand'
+
+export default class InsertNodeFromWorkflowCommand extends InsertNodeCommand {
```
oliver----: Add comment what this is used for

src/article/editor/InsertNodeFromWorkflowCommand.js
```diff
@@ -0,0 +1,12 @@
...
+  execute (params, context) {
+    const workflow = this.config.workflow
+    if (workflow) {
+      context.editor.send('startWorkflow', workflow)
+    } else {
```
oliver----: Explain what the else-branch is used for. ATM it seems not be used. If you do not have good reasons, remove it.

src/article/shared/SupplementaryFileUploadComponent.js
```diff
@@ -0,0 +1,9 @@
+import FileUploadComponent from './FileUploadComponent'
+
+export default class SupplementaryFileUploadComponent extends FileUploadComponent {
```
oliver----: Document what is different to the general FileUpladComponent

oliver----:
We gonna add tests for file upload and reference import in different PRs

PR #1116, opened 2019-01-28
Figure Package Demo

oliver----:
## Why

Polishing the implementation for a Figure Package Demo.

Requirements:
- load example
- can not change the body structure, but otherwise can edit normally
- should be able to edit figure panels as normal

## What

- [x] Add an empty figure-package to see how things look with sparse data
- [x] Supplementary Files not supported well yet in figure legend (there is need for a `<p>` wrapper)
- [x] fix sample: invalid ref in the legend of Figure 1b
- [x] Figure does not rerender when currentPanel is changed (clicking on a thumbnail)
- [x] We should not render empty metadata item for figure panels in metadata mode
- [x] Hghlight current panel in thumbnails
- [x] set node selection when using the panel thumbnails, otherwise selection will be recovered on each update, leading to jumps
- [x] click on drop-down removes DOM selection


Integral:
I sometimes see this picture:
screen shot 2019-01-28 at 22 34 15
this is a bit annoying since you can't understand why the editor is not using visually available space.
The reason is that the current thumbnail list implementation is reserving the whole column on the right of the figure:
screen shot 2019-01-28 at 22 34 01
We did so since we didn't know what to do in case of a big list of thumbnails.

Integral:
I somehow dislike this bold names of custom fields. It's contrasting with title which maybe should dominate. I'm not very opinionated here, just as a remark.
custom_field_names

Issue #1121, opened 2019-01-29
Implement more minimal navigation for sub-figures


michael:
The thumbnails are eating space and cause edge cases (e.g. we need a carousel with size measuring etc.), so we need a more scalable solution. Here it is:

![image](https://user-images.githubusercontent.com/2931/51907660-c3166680-23c7-11e9-84d4-4c74b94e00db.png)

@Integral can you implement this?

michael:
@Melissa37 this is for finishing the current iteration with SourceData. It shows the UI for the authoring view, not for the QC view / author proofing where we have all sub-figures expanded.

oliver----:
Done via #1125

Issue #1120, opened 2019-01-29
Open image in new tab


michael:
Implement a contextual tool that openes the currently selected sub-figure in a new tab. 

This is needed to fully inspect a large image, which is too small when shown in the document flow.

oliver----:
Done via #1125

PR #1125, opened 2019-01-29
Figure package improvements

oliver----:
Working on some improvements for Figure Package:

- [x] Introduce section for Supplementary Files
- [x] enable editing
- [x] implement #1121 
- [x] implement #1120 
- [x] add test for `archive.addAsset()` **(O)**
- [x] add test for open image


oliver----:
@Integral I'd like to want to show you an example of how to improve tests.

this example is good: https://github.com/substance/texture/blob/74152a380c0da0160f5bc5c641c36d8db1267171/test/SupplementaryFile.test.js#L92-L118
It uses helpers so that the test is very 'semantical'. I.e. if we change some UI implementation, most likely it still has the same semantics, and we only need to adapt some helpers. Also it is very easy to read and understand.

This example this is not purely semantical: https://github.com/substance/texture/blob/74152a380c0da0160f5bc5c641c36d8db1267171/test/SupplementaryFile.test.js#L120-L169 
This is much more likely to break. And it is harder to fix, because it takes longer to understand what is done.

We should use comments, and assertion methods, descriptive enough but also as short as possible, following the DRY principle.

oliver----:
Another example of semantic test writing: https://github.com/substance/texture/blob/adf8c05bc344f6d69e6c847e56822690ded5869f/test/Figure.test.js#L246-L283

Issue #1009, opened 2018-10-30
Figures: Cannot save document with figures


pfeiffer0kristina:
I can insert figures, all good, but save the document then does not work, when re-opening the document, figures plus figure references are gone.

tamirhassan:
The same seems to happen to me as soon as there is a bullet list in the content. What's most worrying is that the saving fails silently without returning an error message, so you might not realize before it's too late and you've lost your work.

michael:
We will make sure this is resolved with the upcoming 2.0 release.

@Integral can you try to reproduce this problem and work on a fix?

Integral:
I think so!

PR #1129, opened 2019-01-30
Electron build for Texture 2

Integral:
We should prepare new bundling. 

- [ ] Evaluate and use latest electron
- [ ] Update electron-builder
- [ ] Check the issues with file saving (see #1009)


src/article/editor/DownloadSupplementaryFileTool.js
```diff
@@ -6,7 +6,6 @@ export default class DownloadSupplementaryFileTool extends Tool {
     let el = super.render($$)
     el.append(
       $$('a').ref('link')
-        .attr('target', '_blank')
```
oliver----: We need to do a platform check here, because the browser version does need this.

src/dar/DarFileStorage.js
```diff
@@ -83,6 +83,8 @@ export default class DarFileStorage {
     let { dir, name } = path.parse(darpath)
     // ATTENTION: it is probably possible to create collisions here if somebody uses '@' in a bad way
     // for now, I accepting this because I don't think that this is realistic
+    // dir folder must end with slash since later we will build path again
+    dir += '/'
```
oliver----: It would be better to check if the last character is slash, and only insert if not present, otherwise we could have double slashes

oliver----: Or is it so that path.parse() results in without '/' ? If so, then we should add this to the comment.

src/article/editor/DownloadSupplementaryFileTool.js
oliver----: I have changed and committed this.

src/dar/DarFileStorage.js
oliver----: I checked this. The API says that this is never with '/' so this could should be fine.

oliver----: I have updated the comment accordingly.

oliver----:
@Integral it is important that you keep your branch rebased on the latest master. Otherwise the codecov reports are giving us false alarms.

Issue #1132, opened 2019-02-03
Improve 'Download File' in desktop app.


oliver----:
ATM this opens a link in the same window. This makes the app unusable afterwards.
We should discuss if we need this feature for the desktop app at all, and if so, then we need an explicit mechanism opening a file dialog similar to 'Save as'.

oliver----:
IMO, we should implement the file-dialog mechanism. Even if we decided to disable the save file command in electron, we would probably want something for a future DAR browser.

oliver----:
@Integral what was the reason why you disabled this for electron in first place? https://github.com/substance/texture/blob/master/src/article/editor/DownloadSupplementaryFileTool.js#L12

Integral:
@oliver---- Can I ask you to provide more details regarding application unusability after downloading supplementary file? Is there any error, if so what build did you used, in what OS?
I see that file got downloaded in macOS High Sierra and the app is working afterwards. Still there is a big problem that you have no chance to understand that file got downloaded. True, I introduced this exception for electron since with new window after downloading you'll see a blank window:
<img width="1030" alt="screen shot 2019-02-03 at 14 52 44" src="https://user-images.githubusercontent.com/182010/52176465-2b988580-27c4-11e9-9278-101ac1ab11d4.png">
I think it is rather annoying and confusing. I believe we should try to find some general approach for downloads in desktop app, let's discuss it.

Integral:
@oliver---- thanks for gif!
Can I ask you what branch did you use? Perhaps you can attach a dar file?
Because I don't have this problem with mac os:
![](http://g.recordit.co/GFdHa68wKH.gif)
I not even sure that new window will solve the problem for you. So as a first step I think we should try to understand what's going on.
I also found a problem with attached image as supplementary file. It's not downloading, but opening:
![screen shot 2019-02-03 at 20 11 42](https://user-images.githubusercontent.com/182010/52179824-f4d86480-27ef-11e9-80fa-8a59329a182c.png)

oliver----:
Yes, opening tab is not a solution, but doing in the same tab is worse.

I am on master. Also for me it's opening the file. The one from the capture was very large and it finished after gif ended.

As I have described in the issue description, we need something different: a file save dialog. Similar as our 'Save as'. And instead of opening the tab we need to 'read' data from that url and pass it through to file write.

Integral:
[...]

PR #1134, opened 2019-02-03
Download file improvements

Integral:
## Why

As discussed in #1132 we want to improve file downloading in both browser and electron versions.

## What

For browser version:
  local files with `@download`
  remote with new tab
For electron version:
  show file dialog for saving

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -40,7 +40,11 @@ app.on('ready', () => {
   // Download files
   session.defaultSession.on('will-download', (event, item) => {
     let location = dialog.showSaveDialog({ defaultPath: item.getFilename() })
-    item.setSavePath(location)
+    if (location) {
+      item.setSavePath(location)
+    } else {
+      event.preventDefault()
+    }
   })
 
   createMenu()
--- a/app/main.js
+++ b/app/main.js
@@ -5,7 +5,7 @@ const url = require('url')
 const fsExtra = require('fs-extra')
 
 const {
-  app, dialog, shell, protocol,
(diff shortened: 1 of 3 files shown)


oliver----:
Working now 💯

app/main.js
```diff
@@ -37,6 +37,16 @@ app.on('ready', () => {
     if (error) console.error('Failed to register protocol')
   })
 
+  // Download files
+  session.defaultSession.on('will-download', (event, item) => {
+    let location = dialog.showSaveDialog({ defaultPath: item.getFilename() })
+    if (location) {
+      item.setSavePath(location)
+    } else {
+      event.preventDefault()
```
oliver----: this is the 'Cancel' logic, isn't it? maybe add a comment to make it clear

src/article/editor/DownloadSupplementaryFileTool.js
```diff
@@ -39,6 +38,10 @@ export default class DownloadSupplementaryFileTool extends Tool {
       this.refs.link.el.attr({
         'href': url
       })
+      // Note: in the browser version we want to open remote files in a new tab
```
oliver----: Would be better to have this at the same location in code, as the `@download` thing.

oliver----: I have improved the implementation to be more clear.

app/main.js
Integral: ok, done

oliver----: Thx!

Issue #1147, opened 2019-02-05
pre-populate metadata table with 5 standard rows


evabenitoembo:
Exp system
Method
Measured variable
Perturbations
Disease

michael:
We will add these to the figure package template. Thank you!

oliver----:
Done via #1157

Issue #1136, opened 2019-02-05
Superscript not working in text fields


evabenitoembo:
![image](https://user-images.githubusercontent.com/31538460/52274603-7b568880-294d-11e9-9134-ce536342873a.png)

michael:
We will look into this.

oliver----:
Looks like CSS is missing, or style is broken.

oliver----:
Fixed via #1157

Issue #899, opened 2018-09-19
Add overwrite confirmation dialog


michael:
Currently it's too easy to overwrite existing files when using Texture Desktop > Save/ Save as.

Integral:
@michael @oliver---- I think it's already there in electron@4, look:
![](http://g.recordit.co/5RA9wHk6Km.gif)

Or did you mean something else?

oliver----:
Cool. Then this was easy ;)

Integral:
@oliver---- yes ;) But need something else.

PR #1157, opened 2019-02-06
Desktop app backlog

Integral:
- [x] Update window title when dar file is opened
- [x] Add overwrite confirmation dialog (looks like already there, see [#899](https://github.com/substance/texture/issues/899#issuecomment-460936093))
- [x] Superscript not working in text fields, see #1136
- [x] Pre-populate metadata table with 5 standard rows, see #1147


Integral:
@oliver---- Let me add source data's requests

PR #1164, opened 2019-02-11
App improvements

oliver----:
This PR improves the bundled Texture MacOS app with respect to command line usage.
Moreover, all npm dependencies have been upgraded to latest versions (except `esm` and `nyc`).

On MacOS Texture can be used like this:
```
open -a /Applications/Texture.app MyArticle.dar
```
Additionally, it allows to open an 'inflated' DAR, i.e. the content of the DAR extracted into a folder.
```
open -a /Applications/Texture.app ./Documents/MyArticle/
```


oliver----:
Want to add a test for the storage implementations.

oliver----:
Needed to rollback upgrade of `nyc` and `esm` because the combination of the latest version was creating an empty report, but only on travis, i.e. it was working locally. Might be an npm related problem. Will try again, to upgrade these modules with the next release.

Issue #1166, opened 2019-02-11
Small caps rendering


FAtherden-eLife:
JATS element [sc](https://jats.nlm.nih.gov/archiving/tag-library/1.1d1/n-ary0.html) is used in eLife content to flag small caps rendering. Currently this displays as follows in Texture:

![screen shot 2019-02-11 at 14 15 02](https://user-images.githubusercontent.com/43879983/52568713-71c69800-2e07-11e9-92b1-61bb5104b10c.png)

Markup for reference:
``<abstract>
        <p>This is the abstract. An abstract can contain any formatting, such as <italic>italics</italic>,<bold>bold</bold>,<sup>superscript</sup>,<sub>subscript</sub>or<sc>small caps</sc>. MathML is
                    also allowed:<inline-formula>
            <tex-math>\overset{}{p}_{m} = 0</tex-math>
          </inline-formula>.</p>
      </abstract>``

oliver----:
Fixed via #1169

Issue #1165, opened 2019-02-11
Missing place holder for ref contribs


michael:
![image](https://user-images.githubusercontent.com/2931/52566923-4ba2f700-2e0b-11e9-9dfb-0460cd79ae05.png)

oliver----:
Fixed via #1169

Issue #1156, opened 2019-02-05
substitute ‘sub-figure’ with ‘panel’


evabenitoembo:
![image](https://user-images.githubusercontent.com/31538460/52275859-038a5d00-2951-11e9-97cb-94e25935a3e7.png)

michael:
What if we just call it "Figure"? To avoid the technical term, and I think it's clear based on the context when you are dealing with a group of figures (aka sub-figures).

oliver----:
I'd suggest the following re-wording:
- Add Figure
- Remove Figure
- Move Figure Down
- Move Figure Up
- Open Image
- Replace Image

source-data:
Well, it is an important point and I think it is difficult to re-define the terminology for users. I dont' think it would be clear to most users that we mean "_Panel_" when we write "_Figure_". Users will have very different expectations for "_add a figure_" from "_add a figure panel_" or "_add a sub-panel_".

Figures are a bit special. They are really very very well established elements and many many authors start 'writing' their papers by actually assembling multi-panel figures. Since 90% of figures in the life sciences have panels, there is a rather strong practical justification for being clear and following this de facto standard. 

I know that it is a bit of a pain because for some reason this pervasive structure might not be capture explicitly in JATS under these terms and we have to use "figure groups" as a workaround but this should stay under the hood. For users, I think we should be straighforward and follow the dominant 'behavior'.

oliver----:
I guess we were thinking that 'Panel' is a too technical term. But I am fine with that, too.

source-data:
yes, well, I am open really between Panel and Sub-figure. We can ask users what the prefer. It is just the concept of "figure" that should be clear. We all agree :-)

oliver----:
Done via #1169

Issue #1115, opened 2019-01-28
Internal: add tests for FileUploadComponent


oliver----:
ATM error handling and file-dropping is not tested.

Integral:
Done via #1169.

PR #1169, opened 2019-02-11
Backlog

Integral:
Weekly backlog tasks.

- [x] Missing place holder for ref contribs #1165 
- [x] Small caps rendering #1166
- [x] Underline 
- [x] Overline 
- [x] Strike through 
- [x] Reword ‘Sub-Figure’ actions #1156 
- [x] Add test for FileUploadComponent #1115 
- [x] Add test for ReferenceUpload and DOIImport #1113 
- [x] Show full affiliation in 'Choose Affiliations' pop-up #981
- [x] Ability to add Author and others in Manuscript


oliver----:
@Integral pls rebase to latest master, so you get correct coverage results.

oliver----:
@Integral let's add the tests before merging this. 
I'd rather like to see tests for stuff from past PRs, than implementing a new feature.
Otherwise we end up having no time for writing the tests later, when we have to implement to features.

oliver----:
Let me know if you need any help with these tests.

Integral:
@oliver---- letting you know: I need help with tests.
Also there are problems with rerendering after view switching

oliver----:
@Integral I'll take a look

src/kit/ui/MultiSelectInput.js
```diff
@@ -15,7 +15,7 @@ export default class MultiSelectInput extends OverlayMixin(Component) {
   render ($$) {
     const selected = this.props.selected
     const isEmpty = selected.length === 0
-    const selectedLabels = selected.map(item => item.toString())
+    const selectedLabels = selected.map(item => this.context.api._renderEntity(item))
```
oliver----: I don't like this solution.
MultiSelectInput should be general. So we should rather use renderEntity in the specifc toString() implementation.

oliver----: I will revert this change.

oliver----: Also, it would be good to add a test to 'reveal' the problem before fixing it. This is a general request: Do not just fix things. Let the test-suite detect the problem. Then fix it.
In this case, I can just revert the change, but no effect on the test-suite, and then we are running in circles.

oliver----: @Integral look at my commits to learn how to approach this correctly.

oliver----: Another request: you named the change 'Show more information in affiliations field and popup', which is not what you have done in the change.
It would be better to say 'Use renderEntity() for MultiSelectInput options'.
This would make it very apparent that this is not a desired change.

src/article/shared/AddEntityCommand.js
```diff
@@ -10,9 +10,11 @@ export default class AddEntityCommand extends Command {
     if (workflow) {
       context.editor.send('startWorkflow', workflow)
     } else {
+      // If command requires switching view, then we frst switching to metadate view
+      // and then adding entity to the collection
       const appState = context.appState
       const viewName = appState.get('viewName')
-      if (viewName !== 'metadata') {
+      if (this.config.switchView && viewName !== 'metadata') {
```
oliver----: Why do you need switchView here? Isn't that always the case if viewName !== 'metadata'

oliver----:
This is now as good as we can get currently.

PR #1172, opened 2019-02-15
Multi view history

oliver----:
This PR uses a brand-new Substance feature to allow to retain changes across multiple views.


oliver----:
texture-multi-view-history

PR #1181, opened 2019-02-18
Router

oliver----:
Introduces a router that parses the URL and opens the according location in Texture.
This is particularly useful during development.
But also later we can use this for permanent links.


Integral:
@oliver---- want to discuss this, have small problems here.

Integral:
In short: this will fail now, because of event preventing which I cancelled to be able to change url. There is something behind this event preventing which I don't know.

oliver----:
This is working now.

Issue #1189, opened 2019-02-19
Improve contextual menu UI + mode switcher


michael:
![image](https://user-images.githubusercontent.com/2931/53011398-8c6cc280-3440-11e9-88bd-593c3a80f016.png)

It would improve usability to merge all contextual actions into one menu (displayed on the right edge of the menu bar). We'd also show the category of actions (e.g. Authors, vs Chapters actions).

Additionally I propose to move the mode switcher to the left side, and implement tabs instead of a dropdown (faster switches, and moving all navigation UI closer together).

This would mean the mode switcher is not available in mobile mode, which is acceptable for the moment. However we could have a redundant mode switcher in the regular toolbar, but only show in mobile mode.

oliver----:
Done via #1191

PR #1191, opened 2019-02-19
Ui improvements

Integral:
As suggested in #1189.


oliver----:
This was more massive than we thought this morning.

oliver----:
But I am very proud of it now :)

oliver----:
@Integral how do you like this colour? ;)

oliver----:
@Integral everything seems to be working. Now we need to fix styles. And also still left is hiding inactive groups from the 'Edit' dropdown.

Integral:
👍

Integral:
I would say it is now in a good shape, but want to discuss one tiny detail.

Issue #1187, opened 2019-02-19
Allow to select and edit authors and references from the Manuscript view


oliver----:
Idea: user can click on an author (or reference) which will show a blue border around it, and a contextual tool is presented to edit it. It will then open the specific entity in the metadata view.

oliver----:
Fixed via #1206

PR #1206, opened 2019-02-25
Author and reference editing from manuscript

Integral:
As suggested in #1187 we want to make authors and references selectable and expose contextual tool.
Agenda:
- [x] Introduce author selection
- [x] Introduce reference selection
- [x] Edit author contextual tool
- [x] Edit reference contextual tool

### DIFF
--- a/src/article/ArticleAPI.js
+++ b/src/article/ArticleAPI.js
@@ -132,6 +132,10 @@ export default class ArticleAPI {
     this._setSelection(this._createValueSelection(path))
   }
 
+  selectFirstRequiredPropertyOfMetadataCard (nodeId) {
+    this._setSelection(this._selectFirstRequiredPropertyOfMetadataCard(nodeId))
+  }
+
   _appendChild (collectionPath, data) {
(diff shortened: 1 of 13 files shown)


Integral:
@oliver---- do we want to add more contextual tools now? E.g. create/remove and maybe moving?

Integral:
Almost there with edit commands, but once again have a problem with view switching and scrolling.

Integral:
Ok, looks like we need some other concept for switching view and scroll to. 
There is now `context.articlePanel.refs.content.send('executeCommand', this.name, params)` hack, which is working in case of "always active" commands like entity insertion. But this will not work for contextual tools since command will be disabled in metadata mode and not get executed.
We need to come up with some different idea.

Integral:
Ok, for now we can switch view and scroll to, but without setting a selection.

oliver----:
> @oliver---- do we want to add more contextual tools now? E.g. create/remove and maybe moving?

@Integral let's wait with that for now.

oliver----:
Sweet. Thx!

Issue #1214, opened 2019-03-04
Desktop app don't save after insert figure


yoruyenturk:
Hi,

I tested 2.0 desktop app. We can't save document after insert figure.

michael:
@yoruyenturk can confirm this. Sorry for this. We're working on a fix.

yoruyenturk:
Thanks your response.

Integral:
@yoruyenturk thanks! you can check https://github.com/substance/texture/releases/tag/v2.0.1 now

PR #1217, opened 2019-03-04
Issue 1214

Integral:
As it was reported in #1214, the app can't save document after figure insertion.
This is regression from #1164. FSStorage which is now used only inside node.js via electron RPC calls was calling browser FileReader during Blob -> Array buffer conversion.
For now we will just read file via classical fs.readFile method.

### DIFF
--- a/src/dar/FSStorage.js
+++ b/src/dar/FSStorage.js
@@ -1,13 +1,14 @@
-/* global FileReader, Buffer */
+/* global Buffer */
 import { platform } from 'substance'
 import readArchive from './readArchive'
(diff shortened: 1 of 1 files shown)


Integral:
@oliver---- we will need a node.js test with real image insertion and saving.

oliver----:
@Integral I will put nodejs/electron set of tests for such things on my agenda for the coming weeks.

oliver----:
Regarding the blobs... I need to get my head around this first.
@Integral maybe you can share your experience and insight about the solved problem here to better understand.

Issue #1220, opened 2019-03-08
Allow deletion of references


sbirAECA:
In the new version 2.0.1 it is not enabled the remove item button when I select a reference

**Version: 1.0**

![referenceslist_10](https://user-images.githubusercontent.com/37865559/54016377-f7c5cc80-4182-11e9-95a8-011c255a924a.jpg)

**Version 2.0.1**

![referenceslist_201](https://user-images.githubusercontent.com/37865559/54016305-b6352180-4182-11e9-8a1f-882fe5df1531.jpg)

Thank you

Integral:
@sbirAECA thanks for spotting this regression, this is important one!

Integral:
Looks like remove collection command is always disabled.

sbirAECA:
Not always it works for example when author collections

Integral:
I will investigate. Once again thank you!

Integral:
You are right. Looks like it's just a missing thing. I will add it.

Integral:
This is now fixed in master. We will fix some of other bugs and then do a patch release.
Thanks!

PR #1224, opened 2019-03-08
References deletion

Integral:
## Why

As it was reported in #1220 reference deletion is disabled in a new version of Texture. Indeed we forgot to bring it back.

## What

This PR contains a test revealing a problem, remove reference command and API method for removing references and cleanup xref targets.


oliver----:
Looking good.

Issue #1223, opened 2019-03-08
Footnotes sometimes are in wrong order


Integral:
![](http://g.recordit.co/phnjlqbC2Q.gif)
As far as I know that happens if 
a) if you toggle first occurrence of footnote reference in manuscript
b) you are referencing multiple footnotes there

Integral:
It turns to be a problem with custom labels. So currently if you have a custom footnote label and no xrefs pointing to that footnote it will expose a custom label which is `1` in case of our `kitchen-sink` example.
I've added a test and a fix to #1125.

Integral:
Fixed via #1225.

Issue #1222, opened 2019-03-08
Footnotes removing are not updating xrefs properly


Integral:
![](http://g.recordit.co/YZmc8izdVv.gif)
E.g. if you remove a second footnote you'll see `1,?` xref and will not have a chance to get rid of that question mark.

Integral:
Fixed via #1225.

Issue #1221, opened 2019-03-08
Can't delete table with footnote refs


Integral:
![](http://g.recordit.co/A3Ma1hWBYL.gif)

Integral:
Fixed via #1225.

PR #1225, opened 2019-03-10
Numerous bug fixes

Integral:
Number of bugfixes related to footnotes:
- [x] Footnotes removing are not updating xrefs properly #1222
- [x] Can't delete table with footnote refs #1221
- [x] Footnotes sometimes are in wrong order #1223
- [x] Typo in New > Figure Package menu #1226 
- [x] Figure scoped supplementary file is not exporting #1227

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -284,7 +284,7 @@ function _createMenu () {
               }
             },
             {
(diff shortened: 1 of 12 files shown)


oliver----:
Looking good.

PR #1228, opened 2019-03-18
Custom abstracts

Integral:
- [x] Examples for executive and web summaries
- [x] Introduce Custom Abstract model
- [x] Importer should use first abstract without abstract-type as main abstract and all others abstracts as custom
- [x] Exporter should export main abstract without abstract-type and all others with it 
- [x] Metadata section with abstracts (should show main abstract first and then all custom abstracts)
- [x] Implement CustomAbstractComponent (there you can select abstract type, edit title and abstract)
- [x] Tool for custom abstract insertion 
- [x] Contextual tool to remove a custom abstract
- [x] Contextual tools to move up/down a custom abstract
### Tests:
- [x] Inserting
- [x] Editing title
- [x] Switching type
- [x] Removing
- [x] Reordering


oliver----:
This is looking very good! 👍

PR #1230, opened 2019-03-25
Abstract sections

Integral:
- [x] Add sections to abstract of kitchen-sink (or use a custom abstract for that)
- [x] Allow secs in TextureJATS
- [x] Change abstract.content to allow for headings
- [x] Make sure to use importer for secs for abstract content
- [x] Make sure to use exporter for secs for abstract content
### Tests:
- [x] Switching a paragraph to heading and vice versa


src/article/models/Abstract.js
```diff
@@ -1,6 +1,11 @@
 import { DocumentNode, CONTAINER } from 'substance'
 
-export default class Abstract extends DocumentNode {}
+export default class Abstract extends DocumentNode {
+  getNodes () {
+    const doc = this.getDocument()
+    return this.content.map(id => doc.get(id)).filter(Boolean)
```
oliver----: You should node.resolve() for this

Issue #1219, opened 2019-03-06
Preformat: Not possible to add multiple lines


michael:
It should be possible to edit multi-line preformat blocks. Currently ENTER breaks the node and creates a paragraph underneath. Either we interpret ENTER differently when editing a preformat or at least allow Shift+ENTER to create a new line within the current preformat block.

Integral:
This is not supported by Substance library currently, e.g. [Substance will insert default node type](https://github.com/substance/substance/blob/develop/model/Editing.js#L839-L842).
Newline shift+enter also not supported.

Integral:
I think it would not be a hard thing to change this strategy and make it more configurable. Let's wait for @oliver----.
@michael is that a special thing that we do not use new lines? I think for the `preformat` type or code it is very important feature. Let's look at trivial example:
```
for (var i = 0; i < 9; i++) {
  str = str + i;
}
```
For now you will only be able to input this:
```
for (var i = 0; i < 9; i++) {

  str = str + i;

}
```

oliver----:
This has been solved via #1231

PR #1231, opened 2019-03-27
Fixes

oliver----:
## What ?

- Multiple lines in Preformats #1219: generally solved as `soft-break` feature via `Shift+Enter`. In certain elements `<break>` is used (title, td) and `\n` otherwise
- Bugfix for a problem with updating annotated text properties
- Tidy up


oliver----:
Closing this in favor of #1232

Issue #1243, opened 2019-04-04
Table commands not enabled in Metadata view


oliver----:
**Describe the bug**

In contrast to the Manuscript view, in the Metadata view the context menu for manipulating tables is empty.

**Steps To Reproduce**
- Open Metadata view / Tables
- Select a cell
- Click right mouse button to open context menu

**Expected behavior**

should show a context menu allowing to add/remove rows and columns, for instance

**Screenshots**
![image](https://user-images.githubusercontent.com/284099/55589736-57d37280-5731-11e9-8cb0-67b5c5a0532b.png)

Integral:
@oliver---- would you agree with me that footnotes and bibliographic references inside table must be editable in metadata view as well?

Integral:
...as well as other types of content available for tables in manuscript mode!

oliver----:
Absolutely!

Daniel Beilinson <notifications@github.com> schrieb am Fr., 5. Apr. 2019,
11:34:

oliver----:
Fixed via #1246

Issue #1180, opened 2019-02-18
Link drop-down not available in metadata text fields


FAtherden-eLife:
I am unable to add a link in text fields in the 'Metadata' view. I think this should be possible for text fields. 

If I add a link in the source data then the link text simply displays with underline formatting, with no option to follow the link:
![screen shot 2019-02-18 at 14 13 43](https://user-images.githubusercontent.com/43879983/52956572-69caa300-3387-11e9-98c4-46660d937267.png)

XML:
```<permissions>
        <ali:free_to_read/>
        <license xlink:href="http://creativecommons.org/publicdomain/zero/1.0/">
          <ali:license_ref>http://creativecommons.org/publicdomain/zero/1.0/</ali:license_ref>
          <license-p>This is an open-access article, free of all copyright, and may be freely reproduced, distributed, transmitted, modified, built upon, or otherwise used by anyone for any lawful purpose. The work is made available under the <ext-link ext-link-type="uri" xlink:href="http://creativecommons.org/publicdomain/zero/1.0/">Creative Commons CC0 public domain dedication</ext-link>.</license-p>
        </license>
      </permissions>

oliver----:
Yep. Thanks.

oliver----:
The command for editing links seems not be enabled in Metadata view.

oliver----:
Fixed via #1246

Issue #1158, opened 2019-02-06
data/kitchen-sink/placeholder.png has incorrect extension


diversemix:
The file `data/kitchen-sink/placeholder.png` is being loaded as a png and consequently not shown... even though it is a jpg.
```
$ file data/kitchen-sink/placeholder.png
data/kitchen-sink/placeholder.png: JPEG image data, JFIF standard 1.01, resolution (DPI), density 96x96, segment length 16, Exif Standard: [TIFF image data, big-endian, direntries=1], baseline, precision 8, 1414x1414, frames 3
```

oliver----:
Fixed via #1246

PR #1246, opened 2019-04-05
Backlog

Integral:
This is continuous backlog PR, I will describe what it is solving below.

- [x] Table commands weren't enabled in Metadata view as it was described in #1243
- [x] External link editor wasn't activated for Metadata view as it was described in #1180, plus we didn't render overlay for Metadata view
- [x] Xref editor wasn't activated for Metadata view
- [x] There was a different set of content available for table cells in manuscript and metadata views, so I've added some more in Metadata view:
- inline formula and graphics
- bibliography, figure, formula, footnote, supplementary file, table references
- [x] Cleanup: 
- for the sake of consistency we are now using static getters `refType` in those models which can be referred (we are using them during cross reference insertion commands registration)
- another [LGTM issue](https://lgtm.com/projects/g/substance/texture/rev/fb5e4e01034fb046f70a9ca0406ea66f7acbee47) with never evaluated line of code
- [x] fix incorrect extension of placeholder image from kitchen sink example #1158


src/article/editor/EditorPackage.js
```diff
@@ -209,15 +209,15 @@ export default {
       commandGroup: 'insert'
     })
     config.addCommand('insert-xref-bibr', InsertCrossReferenceCommand, {
-      refType: 'bibr',
+      refType: Reference.refType,
```
oliver----: Yeah :)

oliver----:
Looking good! Thx!

PR #1248, opened 2019-04-10
Plugin framework

oliver----:
## Why

For RDS/Stencila we need to introduce a framework that allows to extend Texture with custom node types, converters and UI components, etc.

## What

This PR refactors the existing implementation generalising the configuration mechanism providing the needed flexibility.
Furthermore, this PR reworks the ArticleLoader with regards to XML validation, transformations, and
custom JATS schemas, which is needed for the RDS/Stencila plugin.


test/CustomAbstracts.test.js
```diff
@@ -28,6 +28,7 @@ test(`Custom Abstracts: add a custom abstract`, t => {
     _addItem(editor, 'custom-abstract')
   }, 'adding new custom abstract should not throw error')
   t.equal(editor.findAll(customAbstractSelector).length, 1, 'there should be one custom abstract')
+  ensureValidJATS(t, app)
```
oliver----: @Integral I'd like you take a look of a new helper I have added in the last PR. Let's add this wherever we think it makes sense, i.e. whenever we have created a bunch of content, this checker makes sure that our exporter is creating valid JATS. Of course this is a very implicit way of testing the exporter, but it is kind of easy to include those checks every once in a while.

Integral: Great, thanks!

Issue #1173, opened 2019-02-18
Move entity renderers to models


Integral:
E.g. what [we did](https://github.com/substance/texture/commit/e832c886abd5772d9f5f2dffcbf2ce16129c6e35) with `Organisation` should be done to other models as well for the sake of consistency.

Integral:
@oliver---- There are two types of renderers here:
1) When you get a plain text string 
2) When you get a markup (for all references)

The first one I moved to member functions, the second pretty complex and I think we should left them as they are for now. Ok?

oliver----:
Sounds good!

Daniel Beilinson <notifications@github.com> schrieb am Do., 18. Apr. 2019,
14:01:

PR #1265, opened 2019-04-18
Dev: move entity renderers to models

Integral:
## Why

See #1173.

## What

Entity renderers for all models except references are moved to member functions.

### DIFF
--- a/src/article/models/Funder.js
+++ b/src/article/models/Funder.js
@@ -2,7 +2,7 @@ import { DocumentNode, STRING } from 'substance'
 
 export default class Funder extends DocumentNode {
   toString () {
-    return this.render()
+    return this.render().join('')
   }
 
   render (options = {}) {
--- a/src/article/models/Funder.js
+++ b/src/article/models/Funder.js
@@ -4,6 +4,17 @@ export default class Funder extends DocumentNode {
   toString () {
     return this.institution
   }
+
+  render (options = {}) {
+    let { awardId, institution } = this
+    let result = [ institution ]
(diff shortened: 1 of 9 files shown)


oliver----:
@Integral we need to find out why this change is decreasing the test coverage.

oliver----:
It seems that the toString() methods are never used.
@Integral do you have a suggestion? or should we just remove the toString() and add them when we need them?

oliver----:
Strange. It must be something more. There are like +84 more missed lines now. I don't see how the toString() would be responsible for this.

oliver----:
@Integral It seems that I have looked at the wrong report. Now it is only +4 misses which maps to the toString() usage. Still, if we are not using them, we should remove them.

oliver----:
Commented out unused toString()
let's bring them in, when we need them.

Integral:
@oliver---- There is a case where we might need something like that, for instance in multi select input, e.g. dropdown with affiliations (see #1266). If it's the only one place where we do this I think for now it's better to do it right in component to keep consistency in methods across entity models.

oliver----:
Just don't want to have dead code from the get go. In case of `Organistation` we are using it already.

oliver----:
An alternative could be to introduce a mixin or super class (RenderableNode) and have the default toString method there.

Integral:
Yes! But let's wait with mixin until we'll sure about that

oliver----:
Then we leave it as is. I.e. without dead code.

Integral:
Absolute! No dead code!

PR #1268, opened 2019-04-24
Keywords (WIP)

Integral:
- [x] Rename property CustomMetadataField.value to ‘values’ and store there array
- [x] Update FigurePanelImporter to map keywords to an array of values
- [x] Update FigurePanelExporter to export the array of values
- [ ] Implement a component for the proposed overlay/popup following the described UX
- [ ] Introduce a command to remove the current value (click on ‘remove’ button should trigger this)
if the cursor is in the removed value, it should be moved to the next ‘reasonable’ place
- [ ] Add a keyboard shortcut to remove the current value (CommandOrCtrl+Delete)
- [ ] Make sure that keyboard navigation via UP/DOWN and TAB/SHIFT-TAB allows to navigate through the different values
- [ ] (O): click on value needs to change selection and/or overlayId so that popup is shown
## Tests:
- [ ] revisit existing tests for keyword editing
- [ ] add tests for the editing of multiple values in a row, i.e. click on value, enter value, ENTER, enter value, ENTER
- [ ] test the value display (i.e. comma separated list of values)
## Mockup:
1QR3hs78dFUY5PiMYqexzU56P06rKKk_0m_w0XGA


oliver----:
@Integral I have implemented this as a prototype, because the required solution is kind of new, while very similar to IsolatedNodes. Let's get everything working, and then see how we could extract a parent class, so that this kind of interface will be easier to implement the next time.

oliver----:
@Integral I tried another approach based on CustomSurface and selections. A little more like IsolatedNodeComponent is working. The implementation is still very hacky.
@michael Before we tidy up, we need to 100% nail down how the UX should be in detail, e.g. keyboard handling, undo/redo, etc.

oliver----:
After playing around with the current interface we came to the conclusion, that the UI feels somewhat unnatural. The reason is not really clear, probably, because it is more like a 'sheet' or 'table' layout, but the UX is not that of a sheet, more like a form. 
We see two options, moving towards a sheet UI, or making the current feel more like a form.
The first, is interesting but too much for this iteration. 

The latter we should give a try, but only as polishing for the current prototype. Potentially a problem with the current design could also be, that the popover/dropdown feel is not appropriate for something more complex like this. We could instead use more space for the popup, maybe more like a little modal, position at the same location, but just let it looks less like a tooltip-like popover.

Refinements:
[ ] Remove the 'delete' buttons and instead add a contextual tool (the buttons clutter the UI too much)
[ ] Try to add a bit more spacing and let the fields look more like form fields
[ ] The popup is moving down when keywords are added

PR #1295, opened 2019-05-10
Overlays

oliver----:
## Why

The current Overlay implementation is made for tools, and does not work for more complex overlay content, including popups with Surfaces in it.

## What

This PR replaces the former Overlay implementation with a generalized OverlayCanvas implementation. All former 'prompt' tools are replaced with components that acquire the OverlayCanvas when necessary.


oliver----:
There is no test triggering usage of `kit/ui/Input` which is only used by `AddSupplementaryFileWorkflow` atm, which seems to be untested.

As this is not the focus of this PR, I will ignore this and add a TODO to `Manuscript.test`

